PR #97, opened 2024-09-20
Remove .vscode/settings.json

cc-a:
Removes vscode settings file and adds to .gitignore.

Having the settings commited requires developers to have a dirty worktree in order to customise project specific settings.


dalonsoa (commented):
In general, the point of having this committed is for the developers contributing to a specific project to have the same settings to improve consistency and NOT customise them at workspace level. If they want customisation, they can do so at User level, right?

cc-a:
Consistency is enforced via CI. I'd argue the developer should be free to have whatever development experience they want in their editor without being forced to configure things at the user level which doesn't have project isolation.

dalonsoa (approved):
OK, then.

jamesturner246 (approved):
No strong opinion here :+1:

alexdewar (commented):
I'm not convinced. I get that it's a bit weird, but it's pretty common practice to commit the `settings.json` file to git (e.g. [here's the one in VS Code's own repo](https://github.com/microsoft/vscode/blob/main/.vscode/settings.json)). It's handy to be able to set the default formatter for Python files etc. Yes, we enforce it properly with pre-commit/CI, but this is really just for convenience.

That said, there are clearly other people out there who want to be able to override the workspace `settings.json` without having a dirty git tree, which seems a sensible proposal: https://github.com/microsoft/vscode/issues/40233

There aren't that many settings in the file, so I don't think it matters too much if we drop it (but I will continue to commit `settings.json` for my own projects :wink:). Up to you!

PR #98, opened 2024-09-20
Add and use custom permissions.

### LINKED ISSUE #12: Placeholder permissions
Eventually we'll need to limit access to different endpoints to only users with the correct permission. Whether or not a user has permission will ultimately need to be determined via the DUNE RBAC system, however this is still currently undefined. As a placeholder we'll hide the views behind an arbitrary permission requirement but all users will have the permission by default.

jamesturner246:
# Description

Adds a set of permissions for modifying processes, and checks users for them when running views.

Tests are added for privileged vs unprivileged clients.

Fixes #12 

## Type of change

## Key checklist

## Further checks


main/models.py
```diff
@@ -1,8 +1,21 @@
...
+    class Meta:
+        """Meta class for the User model."""
+
+        permissions: ClassVar = [
+            ("can_boot_processes", "Can boot processes"),
+            ("can_restart_processes", "Can restart processes"),
+            ("can_flush_processes", "Can flush processes"),
+            ("can_kill_processes", "Can kill processes"),
+            ("can_view_process_logs", "Can view process logs"),
+        ]
```
alexdewar, line 21: Do we need all these separate permissions? I can see you *might* want some users to only be able to view the logs, but does it make sense to have users who can kill and boot processes, but not restart them?

alexdewar (changes requested):
I'm not sure about having so many different permission types, but other than that it all seems good.

main/models.py
jamesturner246: How about one for `view_logs` and one for `modify_processes`?

alexdewar: That makes sense to me: kind of a read vs write distinction.

alexdewar (approved):
LGTM :+1:

cc-a (changes requested):
Implementation looks good. Just needs tests for the views.

dalonsoa (changes requested):
Looks good, but I agree with Chris - we need tests to check that the views have the right `permission_required`.

jamesturner246:
Okay, the test classes now also check for permissions. Ready to go.

cc-a (approved):
Grand. Looks good to me. I'll leave this up for people to look at for now and will probably merge it towards the end of the day.

dalonsoa (approved):
LGTM!

Issue #87, opened 2024-09-19
Select all/deselect all checkbox


cc-a:
Requires #86 first.

Once we have checkboxes for each row we can add a checkbox in the column header that checks/dechecks the checkboxes in all table rows. Behaviour as follows:
- if the header checkbox is checked by the user, all row checkboxes should become checked.
- if the header checkbox is unchecked by the user, all row checkboxes should become unchecked.
- if a row checkbox is toggled resulting in all row checkboxes being checked/unchecked, then the header checkbox should become checked/unchecked.
- if a row checkbox is toggled resulting in a mix of checked and unchecked boxes, then the header checkbox should become indeterminate.

Some experimentation required to figure out the best way to do this. Whilst probably doable with htmx, it may be better to use a scripting solution like [hyperscript](https://hyperscript.org/). At least a little bit of scripting will be required as it is needed to set a checkbox to an indeterminate state.

See #85 for reference.

dalonsoa:
This might come handy, as well: https://stackoverflow.com/a/13053561/3778792

dalonsoa:
I do not understand the reason for handling the undetermined state. That column checkbox is a convenience to check/uncheck all boxes. The backend will use the individual boxes to deal with the logic. From the user perspective, I think it should look "unchecked" if there is a mixture of checked/unchecked boxes.

PR #101, opened 2024-09-24
Add bulk operations for processes

### LINKED ISSUE #86: Bulk operations for restart/kill/flush
Rather than having a column in the process table with restart/kill/flush links for each process, we want to be able to perform operations on multiple processes (i.e. table rows) at once. See #85 for reference.

I believe this can be done by:
- removing the table columns for restart, kill and flush.
- adding a [TemplateColumn](https://django-tables2.readthedocs.io/en/latest/pages/api-reference.html#django_tables2.columns.TemplateColumn) that renders a [checkbox input](https://getbootstrap.com/docs/5.0/forms/checks-radios/#checks) for each row with the process uuid as it's id attribute.
- adding the buttons (as input elements with `type=submit`) for restart, kill and flush.
- wrapping the table + buttons in a form element (don't forget to include the [csrf token](https://docs.djangoproject.com/en/5.1/howto/csrf/)).
- set the [formaction](https://www.w3schools.com/tags/att_input_formaction.asp) attribute for each button.
- Rewriting the view functions for restart, kill and flush to expect POSTed form data and updating the plumbing to drunc to pass multiple uuid's at once.

dalonsoa:
# Description

This PR remove the per-process operations and add checkboxes and buttons such that the operations can be triggered for multiple processes at once. Asa bonus, the code has been simplified significantly.

Fixes #86 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/main/tables.py
+++ b/main/tables.py
@@ -33,6 +33,4 @@ class ProcessTable(tables.Table):
     status_code = tables.Column(verbose_name="Status Code")
     exit_code = tables.Column(verbose_name="Exit Code")
     logs = tables.TemplateColumn(logs_column_template, verbose_name="Logs")
-    restart = tables.TemplateColumn(restart_column_template, verbose_name="Restart")
-    flush = tables.TemplateColumn(flush_column_template, verbose_name="Flush")
-    kill = tables.TemplateColumn(kill_column_template, verbose_name="Kill")
+    select = tables.CheckBoxColumn(accessor="uuid", verbose_name="Select")
--- a/main/templates/main/index.html
+++ b/main/templates/main/index.html
@@ -5,8 +5,14 @@
 
 {% block content %}
 <div class="col">
-  <a href="{% url 'main:boot_process' %}" class="btn btn-primary">Boot</a>
-  {% render_table table %}
+  <form method="post", action="{% url 'main:process_action' %}">
+    {% csrf_token %}
+    <a href="{% url 'main:boot_process' %}" class="btn btn-primary">Boot</a>
+    <input type="submit" value="Restart" class="btn btn-success", name="action">
+    <input type="submit" value="Flush" class="btn btn-warning", name="action">
+    <input type="submit" value="Kill" class="btn btn-danger", name="action">
+    {% render_table table %}
+  </form>
 </div>
 
 {% endblock content %}
--- a/main/templates/main/index.html
+++ b/main/templates/main/index.html
@@ -8,9 +8,9 @@
   <form method="post", action="{% url 'main:process_action' %}">
     {% csrf_token %}
     <a href="{% url 'main:boot_process' %}" class="btn btn-primary">Boot</a>
-    <input type="submit" value="Restart" class="btn btn-success", name="action">
-    <input type="submit" value="Flush" class="btn btn-warning", name="action">
-    <input type="submit" value="Kill" class="btn btn-danger", name="action">
+    <input type="submit" value="Restart" class="btn btn-success", name="action", onclick="return confirm('Restart selected processes?')">
+    <input type="submit" value="Flush" class="btn btn-warning", name="action", onclick="return confirm('Flush selected processes?')">
+    <input type="submit" value="Kill" class="btn btn-danger", name="action", onclick="return confirm('Kill selected processes?')">
     {% render_table table %}
   </form>
 </div>
--- a/main/views.py
+++ b/main/views.py
@@ -112,7 +112,7 @@ def process_action(request: HttpRequest) -> HttpResponse:
         HttpResponse redirecting to the index page.
     """
     action = request.POST.get("action")
-    if action is None:
+    if action is None or action.lower() not in ProcessAction:
         return HttpResponseRedirect(reverse("main:index"))
 
     action_enum = ProcessAction(action.lower())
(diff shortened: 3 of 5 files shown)


dalonsoa:
`kill` and `flush` `drunc` commands can work on multiple processes at once, but `restart` only works for a single process, otherwise there's an error, hence the loop.

main/tables.py
```diff
@@ -33,6 +33,4 @@ class ProcessTable(tables.Table):
     status_code = tables.Column(verbose_name="Status Code")
     exit_code = tables.Column(verbose_name="Exit Code")
     logs = tables.TemplateColumn(logs_column_template, verbose_name="Logs")
-    restart = tables.TemplateColumn(restart_column_template, verbose_name="Restart")
-    flush = tables.TemplateColumn(flush_column_template, verbose_name="Flush")
-    kill = tables.TemplateColumn(kill_column_template, verbose_name="Kill")
+    select = tables.CheckBoxColumn(accessor="uuid", verbose_name="Select")
```
cc-a, line 36: Ah, nice. Didn't know about this. Much better than a template column.

main/templates/main/index.html
```diff
@@ -5,8 +5,14 @@
 
 {% block content %}
 <div class="col">
-  <a href="{% url 'main:boot_process' %}" class="btn btn-primary">Boot</a>
-  {% render_table table %}
+  <form method="post", action="{% url 'main:process_action' %}">
+    {% csrf_token %}
+    <a href="{% url 'main:boot_process' %}" class="btn btn-primary">Boot</a>
+    <input type="submit" value="Restart" class="btn btn-success", name="action", onclick="return confirm('Restart selected processes?')">
```
cc-a, line 11: You can use the [bootstrap_button](https://django-bootstrap5.readthedocs.io/en/latest/templatetags.html#bootstrap-button) template tag to simplify these.

main/views.py
```diff
@@ -80,68 +80,49 @@ class ProcessAction(Enum):
...
     pmd = get_process_manager_driver()
-    query = ProcessQuery(uuids=[ProcessUUID(uuid=uuid)])
+    uuids_ = [ProcessUUID(uuid=u) for u in uuids]
 
     match action:
         case ProcessAction.RESTART:
-            await pmd.restart(query)
+            for uuid_ in uuids_:
+                query = ProcessQuery(uuids=[uuid_])
+                await pmd.restart(query)
```
cc-a, line 97: Why do we need this different logic here? It makes things rather awkward and I'd be tempted to stick with the previous implementation, particularly if we take out the awaits.

tests/test_views.py
```diff
@@ -81,22 +60,34 @@ def test_logs_view_authenticated(self, auth_client, mocker):
...
+        assert response.url == reverse("main:index")
 
-class TestProcessRestartView(ProcessActionsTest):
-    """Tests for the process restart view."""
+    def test_process_action_valid_action(self, auth_client, mocker):
+        """Test process_action view with a valid action."""
+        mock = mocker.patch("main.views._process_call")
+        uuids_ = [str(uuid4()), str(uuid4())]
+        response = auth_client.post(
+            self.endpoint, data={"action": "kill", "select": uuids_}
```
cc-a, line 85: Lets parametrize this to check all expected valid actians.

cc-a (changes requested):
Good stuff. A few tweaks and a question about why the restart command is different.

main/views.py
dalonsoa: `pmd.restart` do not work with more than one `uuid` at the same time, so this one needs a loop. Where we put the loop, I don't mind.

main/templates/main/index.html
dalonsoa: I'm not entirely sure I understand the benefit of this. What is simplified? I still need to add all of those fields, right?

cc-a: Well primarily it handles the styling.

dalonsoa: Not that I don't want to change this, but
1. I'm not sure how to use the tag, and
2. It seems to me that what I use is perfectly fine with bootstrap style, [according to the docs](https://getbootstrap.com/docs/5.0/components/buttons/#button-tags).

cc-a: Well lets leave it for now. See https://github.com/ImperialCollegeLondon/dune_processes/blob/4b7496bc957a32155181e66c4e9277bf25450157/main/templates/registration/login.html#L24 for reference.

cc-a (approved):
Great stuff. Thanks for this.

main/templates/main/index.html
dalonsoa: Yeah, I know how to use the tag in simple cases, but in here I had way more arguments that I don't know how to include in case of the tag.

cc-a: Ah, you just pass them as extra keyword arguments to the tag e.g.:
`{% bootstrap_button button_type="submit" content="Login" onclick="return confirm('Restart selected processes?')" %} `

jamesturner246:
Sorry I'm late! Possible bug with selecting all processes using the top checkbox:

Screenshot from 2024-09-25 21-57-01

I'm guessing this top box is meant to mean 'select all', but when I click it and try e.g. kill, nothing happens.

dalonsoa:
That functionality is meant to be implemented in #87, not here.

jamesturner246:
Ah, I seem No worries. :+1:

PR #106, opened 2024-09-24
Store Kafka messages in user session

### LINKED ISSUE #76: Kafka with user sessions
Now that #7 is complete we have different users with different sessions. We can use a user's session as storage for kafka messages that need to be displayed in the UI. The session is a good place to do this because:

- it lets us filter messages to different users based on topic.
- we only want to store and display messages for a user whilst they are logged in and working i.e. whilst they have an active session.
- the emphemeral nature of session data will take care of tidying up old messages.

cc-a:
# Description

Moves from storing Kafka messages in a hacky global variable to using the Session data of active users. This works by:

- Running the Kafka consumer as a Django admin command. This means it has access to the Django context and settings and can directly populate messages into any active user sessions (i.e. the database).
- Having the index view pop all messages for the session of the user making the request.
- Using appropriate database settings and atomic transactions to prevent race conditions between the consumer process and the web app.

This approach is a fairly basic but gives some nice properties:

- storing messages in the session of each individual user gives an easy mechanism for telling which messages have been displayed to which user.
- the lifetime of messages stored in the sessions is bounded by the standard handling of Django session data (though see below caveat).
- having the consumer populate into the database directly avoids having to think about how to protect an API end point.
- readily supports multiple threads/processes/replicas of the web application.

One caveat is the need to clear the session store in deployment - https://docs.djangoproject.com/en/5.1/topics/http/sessions/#clearing-the-session-store. We should also agree a session expiry period with the project team.

Fixes #76.

## Type of change

## Key checklist

## Further checks


docker-compose.yml
```diff
@@ -2,11 +2,10 @@ services:
   app:
     build: .
     command:
-      - sh
+      - bash
```
alexdewar, line 5: Did plain old `sh` not work?

docker-compose.yml
```diff
@@ -15,14 +14,12 @@ services:
...
-      - KAFKA_URL=kafka:9092
     depends_on:
-      kafka:
-        condition: service_healthy
+      - drunc
   drunc:
     build: ./drunc_docker_service/
     command:
-      - sh
+      - bash
```
alexdewar, line 22: Ditto

main/management/commands/kafka_consumer.py
```diff
@@ -0,0 +1,51 @@
...
+class Command(BaseCommand):
+    """Consumes messages from Kafka and stores them in active user sessions."""
+
+    help = __doc__
+
+    def add_arguments(self, parser: ArgumentParser) -> None:
+        """Add commandline options."""
+        parser.add_argument("--debug", action="store_true")
+
+    def handle(self, debug: bool = False, **kwargs: Any) -> None:  # type: ignore[misc]
```
alexdewar, line 24: Presumably this is because we disallow explicit `Any`? Tbh I'm not sure that setting is super helpful -- there are cases (like this one) where an explicit `Any` makes sense. Maybe we should change that?

main/management/commands/kafka_consumer.py
```diff
@@ -0,0 +1,51 @@
...
+        while True:
+            for messages in consumer.poll(timeout_ms=500).values():
+                message_bodies = []
+                for message in messages:
+                    if debug:
+                        self.stdout.write(f"Message received: {message}")
+                        self.stdout.flush()
+                    bm = BroadcastMessage()
+                    bm.ParseFromString(message.value)
+                    message_bodies.append(bm.data.value.decode("utf-8"))
```
alexdewar, line 41: [Encoding defaults to UTF-8](https://docs.python.org/3/library/stdtypes.html#bytes.decode):

```suggestion
                    message_bodies.append(bm.data.value.decode())
```

alexdewar (approved):
Some v minor suggestions, but otherwise looks good :+1:.

docker-compose.yml
```diff
@@ -48,5 +45,22 @@ services:
       interval: 1s
       timeout: 6s
       retries: 20
+  kafka_consumer:
+    build: .
+    command:
+      - bash
+      - -c
+      - |
+        python manage.py kafka_consumer --debug
```
dalonsoa, line 54: So, just to make sure I get this, we are using the very same code AND database in both apps, but in one we run the server and in the other we run the kafka_consumer. Is that to better use the compute resources or to better manage the different components in case one of them fail?

dalonsoa (approved):
Looks very neat!

docker-compose.yml
cc-a: Not a strictly necessary change in this PR. I found that bash seems to be better at forwarding interrupt signals to child processes so services shutdown more quickly and cleanly.

main/management/commands/kafka_consumer.py
cc-a: I actually like to disallow explicity Any as I think Any's can be a bit of a crutch so prefer that we think carefully about where we use them and explicitly silence the check where necessary.

docker-compose.yml
cc-a: That's right. Ideally Docker Compose is built around the idea of only having one process running in each container. Whilst we could run this is the background under the `app` service we would indeed struggle with handling error states extra. I also think it's cleaner to have the consumer logs in a separate feed.

Ideally we might also run the ssh server in the `drunc` service as a separate service but there is a hard coded restriction in the dummy_boot command that the process must be started on localhost.

cc-a:
Thanks for the reviews so far. I'm planning to merge this towards the end of the day.

main/management/commands/kafka_consumer.py
```diff
@@ -0,0 +1,51 @@
...
+
+                if message_bodies:
+                    with transaction.atomic():
+                        # atomic here to prevent race condition with messages being
+                        # popped by the web application
+                        sessions = Session.objects.all()
+                        for session in sessions:
+                            store = SessionStore(session_key=session.session_key)
+                            store.setdefault("messages", []).extend(message_bodies)
```
jamesturner246, line 50: So `setdefault` handles where `messages` hasn't been added to session, and `extend` is a funny way of saying `append`?

main/management/commands/kafka_consumer.py
```diff
@@ -0,0 +1,51 @@
...
+                        self.stdout.flush()
+                    bm = BroadcastMessage()
+                    bm.ParseFromString(message.value)
+                    message_bodies.append(bm.data.value.decode("utf-8"))
+
+                if message_bodies:
+                    with transaction.atomic():
+                        # atomic here to prevent race condition with messages being
+                        # popped by the web application
+                        sessions = Session.objects.all()
```
jamesturner246, line 47: This gets all *currently logged in* users?

main/management/commands/kafka_consumer.py
jamesturner246, line 1: This command file is in some magic directory which is automatically loaded by Django?

jamesturner246 (commented):
Okay, I follow the logic. Apologies; pile of questions as usual.

Messages are polled from Kafka by a special meta/admin Django function non-stop.

Users *who are logged in at a given moment* have their sessions atomically updated with messages.

This means that if one user pops the messages, they are not popped for all users.

main/management/commands/kafka_consumer.py
cc-a: Right. `setdefault` is shorthand for the below pattern:
```python
value = mydict.get("key")
if value is None:
    value = []
    mydict["key"] = value
```

Extend is basically append but it takes an iterator of values to add rather than a single one. The key thing is that it mutates the existing list.

main/management/commands/kafka_consumer.py
cc-a: It gets all sessions in the database. This broadly corresponds to logged in users subject to the caveat I mention above.

main/management/commands/kafka_consumer.py
cc-a: Correct. Docs - https://docs.djangoproject.com/en/5.1/howto/custom-management-commands/#module-django.core.management.

PR #109, opened 2024-09-25
Kafka message sidebar panel

### LINKED ISSUE #92: Kafka messages in UI
See #85 for reference.

This issue focuses on the layout of the page, behaviours like auto-update and hiding will be considered in separate issues.

Add a side bar to the index page that renders messages sent from Kafka (currently rendered crudely at the top of the index page). The side bar should be the full height of the page and should use the css [overflow-y](https://www.w3schools.com/css/css_overflow.asp) property to provide scolling if content is too large for the screen. You'll need to experiment with the bootstrappy stuff to make things look nice.

jamesturner246:
# Description

Adds a panel on the right to display Kafka messages.

Messages are scrollable when they won't fit. 

I wasn't able to reliably set `max-size` to something that fits the window, so I just set it to 85% viewport height. I think the issue is the main container, which includes the main table, does not yet have its height fixed, so I cannot inherit it's max height minus top bars, padding, etcetera. Is this something you want tackled now, or later?

Else all good to go.

Screenshot from 2024-09-25 16-48-08

Fixes #92 

## Type of change

## Key checklist

## Further checks


main/templates/main/base.html
```diff
@@ -8,6 +8,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     {% bootstrap_css %}
     {% bootstrap_javascript %}
+    {% block extra_css %}{% endblock %}
```
dalonsoa, line 11: Why do we need this empty extra css block here?

dalonsoa (approved):
Looks good to me.

I'd suggest we add [Prettier ](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) code formatter to VScode and the following lines to the settings - for those using it.

```json
  "[html]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode"
  },
```

And also to think on a pre-commit hook for HTML. Now that we have more "complex" HTML, we need to worry about its format as well.

jamesturner246:
Oopp. I think `.vscode/settings.json` is gone now. I'll add it to `.pre-commit-config.yaml` instead.

main/templates/main/base.html
```diff
@@ -1,23 +1,21 @@
-{% load static %}
-{% load django_bootstrap5 %}
-<!DOCTYPE html>
+{% load static %} {% load django_bootstrap5 %}
```
dalonsoa, line 1: I'd suggest to revert running prettrier and add a few tags to tell it to ignore some parts. This headers just look much better if each is int its line. So, leave this as:

```suggestion
<!-- prettier-ignore -->
{% load static %}
{% load django_bootstrap5 %}
```

And the same with the other HTML files. These `{%...` lines are just not understood by prettier as HTML because they are django stuff. Check this commit in Paricia where I've done the change and avoided the reformatting of the django tags. 

https://github.com/ImperialCollegeLondon/paricia/pull/350/commits/8357eeced35cdba04168766c82724d52f22ca995

cc-a:
Perhaps https://github.com/davidodenwald/prettier-plugin-jinja-template

Sent from my T-Mobile 5G Device
Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

alexdewar:
The `prettier` pre-commit hook is kinda dead, which is why we had to change to other hooks :disappointed:.

I'm wondering whether I should revive it though... Don't think it would take much work and I do like the results better than the other hooks. Maybe could be a Hacktoberfest project?

cc-a:
Or https://www.djlint.com/docs/languages/jinja/

jamesturner246:
prettier-plugin-jinja-template looks like it may have the same problem in reverse, where [style tags must be ignored](https://github.com/davidodenwald/prettier-plugin-jinja-template#ignoring-code).

I'll try out djLint. If not I'll just use ignores as in @dalonsoa's comment. 

For now it's reverted.

alexdewar (approved):
LGTM 👍

jamesturner246:
Playing around with djLint and the layout is not bad. Had to add some extra stuff to keep it happy, such as meta fields.

A couple of nags I hid for now: 

* It doesn't like inline CSS, like that in `logs.html`
* A false-positive for an orphaned tag ([upstream bug](https://github.com/djlint/djLint/issues/874))

One final note to factor in is the repo doesn't look especially busy.

cc-a (changes requested):
Good stuff. I like having linting for the templates is a real step forward. I had a play and that orphaned tag one is caused by that stray comma in the form tag.

Curious what others think but because the nature of html leads to lots of indentation I'd prefer to just use 2 spaces instead of 4.

Otherwise this looks like a good first stab at the layout.

jamesturner246:
Stray comma gone. I went ahead and turn indentation down to 2 also.

cc-a (approved):
Nice. I'm afraid the pedant in me can't help but suggest that it would be slightly nicer to put the indent config in pyproject.toml. In case anyone wants to run it outside of pre-commit.

dalonsoa (approved):
Looks good to me - and have stolen the djlint stuff for Paricia :)

Issue #94, opened 2024-09-19
Filter process table


cc-a:
See #85 for reference.

This issues covers only a basic version of this functionality where the user enters data into the filter form and submits it (via dedicated button or hitting return) and the whole page reloads with the filtered data. Auto-updating the table as the user types will be considered in another issue.

Implementation would ideally use [https://django-filter.readthedocs.io/en/stable/index.html](https://django-filter.readthedocs.io/en/stable/index.html) however this seems to only work with data stored in models. We should get a view on #93 first as that may require a move to storing process data in models.

dalonsoa:
To do after #93

dalonsoa:
Let's move on without models.

dalonsoa:
This is what I'll plan do use: https://taskbadger.net/blog/tables.html#bonus-2-filtering

cc-a:
Closed by #153

Issue #89, opened 2024-09-19
Process table auto-refresh


cc-a:
Depends on #88.

Should be relatively easy to accomplish by [polling with htmx](https://htmx.org/docs/#polling):

- Add a view function that renders just the process table template.
- Add a div wrapper to the table in the index view that polls the process table view and replaces it's inner html to update the page.
- Make sure that the url that is polled by htmx includes any query parameters present in the url. This will ensure that e.g. table sorting is preserved.

See #85 for reference.

PR #130, opened 2024-10-03
Htmx setup and process table polling

### LINKED ISSUE #88: HTMX setup
- Download the latest version of the htmx javascript and commit it in `main/static/main/js/`.
- Add a script tag in the header of the base template that loads the library.
- Refactor templates to easily support partial loading of page elements.

Going into detail on the last point. The approach used by htmx means we need to be able to use the Django templating engine to render parts of the page in isolation as well as those parts as part of a larger template. I suggest the following structure to do this and it's application to the process table of the index view as a proof of principle:

- Create the directory `main/templates/main/partials/` which will be used store page elements that need to be renderable individually as well as part of a page.
- Create `main/templates/main/partials/process_table.html` and move the table rendering logic from the index page template into it.
- Update `main/templates/main/index.html` with `{% include 'main/partials/process_table.html' %}`.

cc-a:
# Description

We covered this in detail in the session yesterday but this PR:

- Adds the htmx javascript library as a static file and loads it in the base template.
- Refactors templates and views so that the index page and process table are rendered in seperate views/templates.
- Adds htmx functionality to the index view to perform initial loading of the process table.
- Adds htmx functionality to the table view to use [load polling](https://htmx.org/docs/#load_polling) to update the table.

This implementation supports the checkbox column and retaining which checkboxes are clicked on update. It does not currently support table sorting so this has been disabled. As filtering (#94) is planned for the table we should check if the project team is that bothered about sorting before we implement it.

The implementation we looked at yesterday used only the `load` trigger in the process_table template to update the table. This left open a race condition where checking a box after a polling request was made but before the response was received meant the table was not updated correctly. To avoid this I've added the `click from:input` trigger. This means that an update is triggered whenever one of the boxes is toggled. This request created by the click supersedes any polling request currently in flight keeping the table in sync.

Fixes #88  and #89.

## Type of change

## Key checklist

## Further checks


dalonsoa (approved):
Looks ok! I've just had a couple of questions, but the tutorial yesterday was good to explain all.

AdrianDAlessandro (approved):
LGTM!

main/templates/main/partials/process_table.html
```diff
@@ -0,0 +1,4 @@
+{% load render_table from django_tables2 %}
+<div hx-post="{% url 'main:process_table' %}"
+     hx-trigger="load delay:1s, click from:input"
+     hx-swap="outerHTML">{% render_table table %}</div>
```
jamesturner246, line 4: So `render_table` is substituted immediately on page load with a bootstrap div. Then here the new div's trigger fires 1 sec after load, substituting into itself, re-rendering the table and restarting the trigger timer. `process_table` is in its own page allowing HTMX to just call it, rather than `index`. Is `hx-post` used to send checkbox state, rather than `hx-get`?

jamesturner246 (approved):
LGTM :+1:

Sahil590 (approved):
LGTM

PR #131, opened 2024-10-03
Rename project to drunc_ui

AdrianDAlessandro:
# Description

This PR changes the Django project name to `drunc_ui`. It will also require a change to the GitHub repository name to go along with it. This should not cause problems with your local remotes as GitHub will set up redirects from the old name, but they still recommend changing those remotes. [See GitHub docs](https://docs.github.com/en/repositories/creating-and-managing-repositories/renaming-a-repository). I've tagged everyone involved for review so they know to change their local remotes.

Once we are happy with this, I will rename the GitHub repo. I believe the link checker workflow will fail until I do that.

Fixes #112 

## Type of change

- [ ] Documentation (non-breaking change that adds or improves the documentation)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (non-breaking, back-end change that speeds up the code)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Breaking change (whatever its nature)

## Key checklist

- [x] All tests pass (eg. `python -m pytest`)
- [ ] The documentation builds and looks OK (eg. `python -m sphinx -b html docs docs/build`)
- [x] Pre-commit hooks run successfully (eg. `pre-commit run --all-files`)

## Further checks

- [ ] Code is commented, particularly in hard-to-understand areas
- [ ] Tests added or an issue has been opened to tackle that in the future. (Indicate issue here: # (issue))

Issue #113, opened 2024-09-26
Refactor main app


cc-a:
Whilst we're developing both the controller and process manager UI's under this application we'd like to keep the door open to split them at a later date. To support this we should break up the `main` app such that:

- all models/views common to both the controller and process manager UI should be in a new Django app (called `common` or something like that) e.g. User model, authentication views. Don't forgot the update the [view namespace](https://github.com/ImperialCollegeLondon/dune_processes/blob/a34e2c9108247eab87b92715acadaae70d3129b0/main/urls.py#L7).
- all process manager related functionality should be in a new Django app. As above make sure views are given a dedicated namespace.
- all process manager views should be moved under the path prefix `/process_manager/`
- split view tests into separate files and put in appropriate sub directories.

PR #132, opened 2024-10-03
Initialise the controller app

### LINKED ISSUE #114: Initialise controller app
To hold the controller functionality. 

See the Django tutorial for a outline: https://docs.djangoproject.com/en/5.1/intro/tutorial01/#creating-the-polls-app.

Make sure:

- there is a placeholder index view with a test.
- all view functions are namespaced under `controller` or similar (as in the [main module](https://github.com/ImperialCollegeLondon/dune_processes/blob/a34e2c9108247eab87b92715acadaae70d3129b0/main/urls.py#L7)).
- urlpatterns are included in the project `urls.py` with an appropriate prefix (e.g. `/controller/`).

AdrianDAlessandro:
# Description

This PR adds a simple controller app to the repo. All it has is an index view and is linked to the current project in INSTALLED_APPS. I have tried to make as few changes as possible so that it will not have conflicts with #131 and when #113 is tackled.

The first commit is what is auto-generated by `python manage.py startapp controller`. So if you want to see my customisations, ignore that commit.

Fixes #114 

## Type of change

## Key checklist

## Further checks


tests/utils.py
jamesturner246, line 1: Can this go in conftest.py?

jamesturner246 (approved):
Looks good. :+1: 

Tiny tests question.

tests/utils.py
AdrianDAlessandro: Yes, but it would still need to be imported into each test file with `from .conftest import LoginRequiredTest`. And considering contest is automatically included, it felt more sensible to put it into a separate file. But I'm not sold on that choice.

dalonsoa (approved):
This looks ok, however, as it is done, it seems that the current app (the process manager) is still the main app and the controller a dependent, somehow, based on the URL patterns and that the template is the one of the main app. 

Possibly this is part of another PR, but I think that `main` should be setup as an app in parallel to the controller, with all the common templates in the root directly and the URLs symmetric, one `process_manager/` and the other `controller/`.

AdrianDAlessandro:
> This looks ok, however, as it is done, it seems that the current app (the process manager) is still the main app and the controller a dependent, somehow, based on the URL patterns and that the template is the one of the main app.
> 
> Possibly this is part of another PR, but I think that `main` should be setup as an app in parallel to the controller, with all the common templates in the root directly and the URLs symmetric, one `process_manager/` and the other `controller/`.

Yes, that is to be tackled in #113

tests/utils.py
cc-a: Having the separate utils.py is the right call I think. conftest.py is magic pytest nonsense. This is just code we want to reuse.

cc-a (approved):
Good stuff.

PR #133, opened 2024-10-04
Create process_manager app

### LINKED ISSUE #113: Refactor main app

AdrianDAlessandro:
# Description

This PR creates a new app called `process_manager` and shifts all the behaviour for the process manager from the `main` app to the new app. The only details remaining in the main app are:
- The user model (plus associated admin, migrations, and login template)
- The base and navbar templates
- The Kafka consumer management command

Some more details:
- The `main` app index page still requires a login, all it shows is two buttons to navigate to the other apps from.
- The `process_manager` views and urls are namespaces under the path prefix `/process_manager/`
- The tests have been split into sub-folders for each app
- The navbar template is newly separated out and has additional nav links to the two apps, which can only be viewed when authenticated

Caveat:
- It's possible some of the behaviours currently in the `process_manager` will end up being common to the `controller`, but this information is not yet known for sure. If that occurs in the future, we can simply shift things back to the `main` app ass needed.

This PR builds on #132 and should be merged after that.
Close #113 

## Type of change

- [x] Breaking change (whatever its nature)

## Key checklist

## Further checks


cc-a (approved):
LGTM. htmx stuff still seems to be working.

PR #138, opened 2024-10-07
General refactor

cc-a:
# Description

This PR takes a pass at refactoring (mostly moving stuff around) the code base and simplifying some of the tests. It:

- Separates the view functions in `process_manager/views.py` into 3 submodules handling pages (full individual pages), partials (elements embedded within larger pages used by HTMX) and actions (acting on dune processes). The tests are also split accordingly.
- Pulls out commands that interact with the process manager into `process_manager/process_manager_interface.py` and standardises how we wrap async calls to `ProcessManagerDriver` methods.
- Reduces the use of adhoc mocking by adding an auto-used mock fixture that prevents gRPC calls being made during tests.
- Shortens the names of tests that are class methods to reduce redundant information.
- Adds a `PermissionRequired` test class as an extension to `LoginRequired` which checkes that an authenticated but unpriveleged user is blocked from a view correctly.
- Neatens some of the fixtures a bit.

## Type of change

## Key checklist

## Further checks


process_manager/views/pages.py
```diff
@@ -0,0 +1,69 @@
...
+@login_required
+def index(request: HttpRequest) -> HttpResponse:
+    """View that renders the index/home page."""
+    with transaction.atomic():
+        # atomic to avoid race condition with kafka consumer
+        messages = request.session.load().get("messages", [])
+        request.session.pop("messages", [])
+        request.session.save()
+
+    context = {"messages": messages}
```
alexdewar, line 26:
```suggestion
    context = dict(messages=messages)
```

tests/process_manager/views/__init__.py
alexdewar, line 1: I don't think you actually need `__init__.py` files in test folders. Won't hurt though.

process_manager/views/actions.py
```diff
@@ -0,0 +1,31 @@
...
+        request: Django HttpRequest object.
+
+    Returns:
+        HttpResponse redirecting to the index page.
+    """
+    try:
+        action = request.POST.get("action", "")
+        action_enum = ProcessAction(action.lower())
+    except ValueError:
+        return HttpResponseRedirect(reverse("process_manager:index"))
```
alexdewar, line 27: What is it that might raise `ValueError` here? Is it `POST.get()`?

process_manager/process_manager_interface.py
```diff
@@ -0,0 +1,103 @@
...
+    Args:
+      uuid: UUID of the process.
+
+    Returns:
+      The process logs.
+    """
+    return asyncio.run(_get_process_logs(uuid))
+
+
+async def _boot_process(user: str, data: dict[str, str | int]) -> None:
```
alexdewar, line 90: In a similar vein, I guess `data` could be a `Mapping`.

alexdewar (approved):
Looks good! I've made some small comments.

Not really related to this PR: but it feels kinda wrong that we're wrapping all the `async` calls to make them synchronous. The DUNE folks have given us a nice `async` API and it seems a pity to circumvent it! It looks like Django can handle async views ([see here](https://docs.djangoproject.com/en/5.1/topics/async/), with the caveat that if your middleware isn't async-aware then it'll just fall back on using threads internally. In any case, if we could do things that way we could avoid some of the boilerplate wrapping code. Worth opening an issue for?

tests/process_manager/views/__init__.py
cc-a: I think we do in this case otherwise the relative import of the `tests/utils.py` doesn't work.

process_manager/views/actions.py
cc-a: I believe it's `ProcessAction` if `action.lower()` is not a valid value for the enum.

alexdewar: Ok cool. Might be helpful to have a comment saying that?

cc-a: Done.

process_manager/process_manager_interface.py
```diff
@@ -0,0 +1,104 @@
...
+
+async def _get_session_info() -> ProcessInstanceList:
+    pmd = get_process_manager_driver()
+    query = ProcessQuery(names=[".*"])
+    return await pmd.ps(query)
+
+
+def get_session_info() -> ProcessInstanceList:
+    """Get info about all sessions from process manager."""
+    return asyncio.run(_get_session_info())
```
AdrianDAlessandro, line 34: Is this what you meant by standardising how async methods are wrapped? Since this is a pattern that's re-used a number of times, it feels like maybe it's better suited to an OOP approach. But I think leaving it like this for now is fine until we make a decision on #140

tests/conftest.py
```diff
@@ -14,29 +14,40 @@ def auth_client(django_user_model) -> Client:
...
+    permission = Permission.objects.get(codename=permission_name)
     user.user_permissions.add(permission)
     client = Client()
     client.force_login(user)
     return client
 
 
+@pytest.fixture
+def auth_process_client(django_user_model) -> Client:
+    """Return a authenticated client with modify process privilege."""
```
AdrianDAlessandro, line 28:
```suggestion
    """Return an authenticated client with modify process privilege."""
```

tests/conftest.py
```diff
@@ -14,29 +14,40 @@ def auth_client(django_user_model) -> Client:
...
+    return _privileged_user_client(
+        django_user_model, "process_user", "can_modify_processes"
+    )
+
+
 @pytest.fixture
 def auth_logs_client(django_user_model) -> Client:
     """Return a authenticated client with view logs privilege."""
```
AdrianDAlessandro, line 36:
```suggestion
    """Return an authenticated client with view logs privilege."""
```

tests/conftest.py
```diff
@@ -14,29 +14,40 @@ def auth_client(django_user_model) -> Client:
...
 def dummy_session_data() -> dict[str, str | int]:
     """A dictionary of dummy data to populate a dummy session."""
     return dict(session_name="sess_name", n_processes=1, sleep=5, n_sleeps=4)
+
+
+@pytest.fixture(autouse=True)
+def grpc_mock(mocker):
+    """Mock out the method that generates gRPC calls to external interfaces."""
+    yield mocker.patch(
+        "process_manager.process_manager_interface.ProcessManagerDriver.send_command_aio"
```
AdrianDAlessandro, line 52: Mocking specifically the part that generates the gRPC calls like this is good. We should now be able to add some tests that check the views are calling the correct methods of the ProcessManagerDriver

tests/test_process_manager_interface.py
AdrianDAlessandro, line 1: We should open an issue to expand this test module to get more coverage of the `process_manager_interface`

tests/process_manager/views/test_action_views.py
```diff
@@ -0,0 +1,42 @@
...
+        response = auth_process_client.post(
+            self.endpoint, data={"action": "invalid_action"}
+        )
+        assert response.status_code == HTTPStatus.FOUND
+        assert response.url == reverse("process_manager:index")
+
+    @pytest.mark.parametrize("action", ["kill", "restart", "flush"])
+    def test_valid_action(self, action, auth_process_client, mocker):
+        """Test process_action view with a valid action."""
+        mock = mocker.patch("process_manager.views.actions.process_call")
```
AdrianDAlessandro, line 34: Is this patch no longer needed now because of the auto-used mock fixture? The `mock.assert_called_once_with` can still be used if we use a `mocker.spy`

tests/process_manager/views/test_page_views.py
```diff
@@ -0,0 +1,101 @@
...
+        with assertTemplateUsed(template_name="process_manager/index.html"):
+            response = auth_client.get(self.endpoint)
+        assert response.status_code == HTTPStatus.OK
+
+    def test_admin(self, admin_client):
+        """Test the index view for an admin user."""
+        with assertTemplateUsed(template_name="process_manager/index.html"):
+            response = admin_client.get(self.endpoint)
+        assert response.status_code == HTTPStatus.OK
+        assertContains(response, "Boot</a>")
```
AdrianDAlessandro, line 26: This test doesn't add anything, if you replace `admin_client` with `auth_client` it will still pass.

tests/process_manager/views/test_page_views.py
```diff
@@ -0,0 +1,101 @@
...
+
+class TestLogsView(PermissionRequiredTest):
+    """Tests for the logs view."""
+
+    uuid = uuid4()
+    endpoint = reverse("process_manager:logs", kwargs=dict(uuid=uuid))
+
+    def test_get(self, auth_logs_client, mocker):
+        """Test the logs view for a privileged user."""
+        mock = mocker.patch("process_manager.views.pages.get_process_logs")
```
AdrianDAlessandro, line 55: Also, is this patch no longer needed?

tests/process_manager/views/test_page_views.py
```diff
@@ -0,0 +1,101 @@
...
+            response = auth_process_client.post(
+                reverse("process_manager:boot_process"), data=dict()
+            )
+        assert response.status_code == HTTPStatus.OK
+
+        assert "form" in response.context
+
+    def test_post_valid(self, auth_process_client, mocker, dummy_session_data):
+        """Test the POST request for the BootProcess view."""
+        mock = mocker.patch("process_manager.views.pages.boot_process")
```
AdrianDAlessandro, line 93: Also, is this patch no longer needed?

AdrianDAlessandro (approved):
This is good, I've been wanting to do the same kind of refactor. I have a lot of comments, but they're a combination of questions and reminders to open new issues. So I'm not requesting changes.

There are some missing docstrings that don't seem to be getting caught by linting (I think they're all functions that begin with "_"). Is this intentional?

drunc_ui/settings/settings.py
```diff
@@ -151,3 +153,5 @@
 CRISPY_TEMPLATE_PACK = "bootstrap5"
 
 KAFKA_ADDRESS = os.getenv("KAFKA_ADDRESS", "kafka:9092")
+
+django_stubs_ext.monkeypatch()
```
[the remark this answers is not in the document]
cc-a: The docs in the django-stubs repository cover it for a related use case - https://github.com/typeddjango/django-stubs?tab=readme-ov-file#i-cannot-use-queryset-or-manager-with-type-annotations. Adding it here lets us correctly type the `BootProcessView` class and remove the mypy ignore that was there previously.

process_manager/process_manager_interface.py
```diff
@@ -0,0 +1,104 @@
...
+def get_process_manager_driver() -> ProcessManagerDriver:
+    """Get a ProcessManagerDriver instance."""
+    token = create_dummy_token_from_uname()
+    return ProcessManagerDriver(
+        settings.PROCESS_MANAGER_URL, token=token, aio_channel=True
+    )
+
+
+async def _get_session_info() -> ProcessInstanceList:
+    pmd = get_process_manager_driver()
```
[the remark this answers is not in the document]
cc-a: Things starting with an underscore aren't checked on the assumption that they're private. In this case I skipped it as these functions directly map to a public function in the module that is documented.

tests/process_manager/views/test_action_views.py
cc-a: Interesting. I didn't know about that.

tests/process_manager/views/test_page_views.py
cc-a: I agree. Removed.

process_manager/process_manager_interface.py
jamesturner246, line 1: Okay, so the `get_whatevers` are just placeholders until async views are done/decided on.

tests/conftest.py
jamesturner246: Not sure where/how this is used.

jamesturner246 (approved):
Still following. Looks good to me :+1:

process_manager/views/pages.py
```diff
@@ -0,0 +1,69 @@
...
+    return render(
+        request=request, context=context, template_name="process_manager/logs.html"
+    )
+
+
+class BootProcessView(PermissionRequiredMixin, FormView[BootProcessForm]):
+    """View for the BootProcess form."""
+
+    template_name = "process_manager/boot_process.html"
+    form_class = BootProcessForm
```
dalonsoa, line 55: Why do we need `BootProcessForm` in when indicating the class to inherit from? It doesn't seem to add much to aid the type checking, as far as I can tell.

process_manager/views/pages.py
```diff
@@ -0,0 +1,69 @@
...
+    success_url = reverse_lazy("process_manager:index")
+    permission_required = "main.can_modify_processes"
+
+    def form_valid(self, form: BootProcessForm) -> HttpResponse:
```
dalonsoa, line 59: There's nothing in this method that uses specific features from the `BootProcessForm`, so using just `forms.Form` is more generic and enough for the type checker to work.
```suggestion
    def form_valid(self, form: form.Form) -> HttpResponse:
```

process_manager/views/partials.py
```diff
@@ -0,0 +1,54 @@
...
+
+@login_required
+def process_table(request: HttpRequest) -> HttpResponse:
+    """Renders the process table.
+
+    This view may be called using either GET or POST methods. GET renders the table with
+    no check boxes selected. POST renders the table with checked boxes for any table row
+    with a uuid provided in the select key of the request data.
+    """
+    selected_rows = request.POST.getlist("select", [])
```
dalonsoa, line 23: When it is a GET method, does POST still exist with no `select`? Otherwise, how do we distinguish between a GET and a POST?

tests/conftest.py
dalonsoa: This is used automatically before every test is run - see the `fixture(autouse=True)` so we don't need to worry about mocking the calls to `gRPC` in every test that might use them.

dalonsoa (approved):
I have a couple of questions, but otherwise it looks good to me!

process_manager/views/pages.py
cc-a: I think that `BootProcessForm` makes the most sense as that is the type we expect in order for the correct data to be present for the function to work correctly. If it was just typed as `Form` then any `Form` subclass would be valid even though these don't contain the required data.

process_manager/views/pages.py
cc-a: `FormView` is a generic type and we're disallowing Any generics so we either have to provide this or ignore the error.

process_manager/views/partials.py
cc-a: It does still exist. We make use of the case that there is no difference in behaviour of the function when there is a GET request or a POST request with no data.

PR #139, opened 2024-10-07
Include all relevant modules in test coverage.

cc-a:
# Description

On reflection I thought our 100% test coverage looked a little optimistic and turns out we've been excluding the modules with most of the code. Adds `process_manager` and `controller` modules.

Fixes # (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,7 +56,7 @@ ignore_missing_imports = true
 django_settings_module = "drunc_ui.settings"
 
 [tool.pytest.ini_options]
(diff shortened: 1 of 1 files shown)


alexdewar (approved):
LGTM. Would it also work to just include coverage for everything (i.e. the root dir) and then just exclude the few files we don't want checked?

cc-a:
> LGTM. Would it also work to just include coverage for everything (i.e. the root dir) and then just exclude the few files we don't want checked?

As I forgot to include the main module I think this is a good idea. Better to default include things we don't want than to miss things we do.

AdrianDAlessandro (approved):
Good idea! Something I missed when splitting the main app into the separate apps

jamesturner246 (approved):
LGTM :+1:

PR #143, opened 2024-10-09
Toggle

### LINKED ISSUE #87: Select all/deselect all checkbox

dalonsoa:
# Description

Adds functionality to the checkbox in the table header such that when checked/unchecked, the others behave accordingly. Likewise, if all are checked and one of the checkboxes is unchecked, then the header box is also unchecked. 

It uses a combination of JS in the template and Python in the partial view. The best I could do due to the HTML refresh.

~~The actual functionality works - just try it - but the test created for that purpose fails because, inexplicably, the `table_data` object within the partial view is empty. No idea what's wrong there, so any suggestion is most welcomed. The test has been marked as `xfail` for now.~~

Fixes #87

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/main/templates/main/base.html
+++ b/main/templates/main/base.html
@@ -15,6 +15,8 @@
     {% bootstrap_javascript %}
     {% block extra_css %}
     {% endblock extra_css %}
+    {% block extra_js %}
+    {% endblock extra_js %}
     <script src="{% static 'main/js/htmx.min.js' %}"></script>
     <script src="{% static 'main/js/_hyperscript.min.js' %}"></script>
   </head>
--- a/process_manager/tables.py
+++ b/process_manager/tables.py
@@ -41,5 +41,8 @@ class ProcessTable(tables.Table):
     exit_code = tables.Column(verbose_name="Exit Code")
     logs = tables.TemplateColumn(logs_column_template, verbose_name="Logs")
     select = tables.CheckBoxColumn(
-        accessor="uuid", verbose_name="Select", checked="checked"
+        accessor="uuid",
+        verbose_name="Select",
+        checked="checked",
+        attrs={"th__input": {"onclick": "toggle(this)"}},
     )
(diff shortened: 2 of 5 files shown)


jamesturner246 (commented):
All the behaviours are there. Tested:

* 'all' box toggles all boxes on/off
* manual click all on/off and 'all' box is respectively on/off
* all box behaves when some boxes manually selected (on/off)

One thing I noticed was it sometimes ignored my click, or flashed quickly on and off. On Firefox, if that makes any difference.

cc-a:
I'm don't totally understand why but you can get all tests to pass by changing test_partial_views.py:L33 to `mock().data.values.__iter__.return_value = instance_mocks`.

dalonsoa:
Well, I've no idea, but it does work.

@jamesturner246 , I don't know why that might be. In Edge it seems quite smooth.

cc-a (approved):
Looks good. In principle I'd prefer we either picked vanilla javascript or hyperscript for tweaking client side behaviours but I'd like to take another pass at how we're handling the checkboxes in general to avoid having to send the data to the server all the time.

jamesturner246:
> @jamesturner246 , I don't know why that might be. In Edge it seems quite smooth.

I was spamming clicks very quickly, to be fair!

Issue #96, opened 2024-09-19
Add logged in user info


cc-a:
See #85 for reference. This refers to the information in the top right showing the username of the currently logged in user.

Issue #91, opened 2024-09-19
Help button


cc-a:
See #85 for reference.

Add a help button to the UI that opens a standalone page with some placeholder text.

PR #145, opened 2024-10-09
adding user info and help page setup

Sahil590:
# Description
Added the current user login info in the Navar.
Initial setup of the help page.
 

Fixes # (issue)
#91 
#96 
## Type of change

## Key checklist

## Further checks

### DIFF
--- a/main/templates/main/help.html
+++ b/main/templates/main/help.html
@@ -1,7 +1,4 @@
 {% extends "main/base.html" %}
-{% block title %}
-  Help
-{% endblock title %}
 {% block content %}
   <h1>Help</h1>
   <p>This is the help page.</p>
--- a/main/templates/main/help.html
+++ b/main/templates/main/help.html
@@ -1,0 +1,8 @@
+{% extends "main/base.html" %}
+{% block title %}
+  Help
+{% endblock title %}
+{% block content %}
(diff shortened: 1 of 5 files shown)


main/templates/main/help.html
```diff
@@ -0,0 +1,8 @@
+{% extends "main/base.html" %}
+{% block title %}
+  Help
+{% endblock title %}
```
dalonsoa, line 4: I've seen it in other pages - I think the process manager and controller do the same - but why do we have a title block if we are then adding a `h1` header within the content block?

dalonsoa (approved):
LGTM!

main/templates/main/help.html
Sahil590: This was just something I was playing around with, wanted to see the difference between the two.

cc-a: The title block doesn't add content to the page it sets the content of the [title tag](https://www.w3schools.com/tags/tag_title.asp) within the html header - https://github.com/ImperialCollegeLondon/drunc_ui/blob/33b8433fabd19ccf6ae7e2f05c2f674463b5dbc0/main/templates/main/base.html#L6-L10. This effectively sets the name of the tab in the browser.

PR #146, opened 2024-10-10
Add Kafka message model

### LINKED ISSUE #119: Add database model for messages
The current session based implemention for messages is insufficient to support required functionality - most notably persistence on refresh.

- Create a BroadcastMessage model class that stores message text and timestamp.
- Update `kafka_consumer` admin command to populate messages into the BroadcastMessage table and to prune old messages from the list.
- On index page load, populate the Kafka message feed with all (or a default time window) of saved messages.

Not considered part of this issue but this implementation will also support auto-population of the message feed in the UI as it can support polling for messages newer than a given timestamp.

jamesturner246:
# Description

Add and use a new table model for persistent Kafka messages. I've named it `DruncMessage`, since `BroadcastMessage` is already taken by the druncschema package.

not sure if it's worth passing in timestamps separately to the view, so we can format them nicer. Something like [this](https://getbootstrap.com/docs/5.2/components/list-group/#custom-content).

Currently a temp `MESSAGE_EXPIRE_SECS = 60` to control this, so old messages begin purging after 60 seconds.

Pondering: is there ever a danger of `id` overflowing? How many messages are going to be chucked out here, over the whole period that the system is running?

Also the Django command only seems to run when in the 'process manager' page -- i.e. no stale deleting occurs unless we are on that page.

Fixes #119 

## Type of change

## Key checklist

## Further checks


cc-a (changes requested):
Great. Just a few tweaks needed I think.

I'm not worried about the overflow if we ran the maths I think we'd probably find we have a few centuries to play with.

One thing that does give me pause is that under realistic scenarious there might be thousands of messages in the database so re-rendering all of them every second for multiple users might start to get noticeable on the server side. I think we should create an issue to look at some load testing and performance in deployment so we keep it in mind.

main/management/commands/kafka_consumer.py
```diff
@@ -31,21 +32,46 @@ def handle(self, debug: bool = False, **kwargs: Any) -> None:  # type: ignore[mi
...
+                        # displayed by the web application
+                        DruncMessage.objects.bulk_create(
+                            [
+                                DruncMessage(timestamp=t, message=msg)
+                                for t, msg in zip(message_timestamps, message_bodies)
+                            ]
+                        )
+
+                # TODO: handle this (and timezone?) via settings.py
+                MESSAGE_EXPIRE_SECS = 60
```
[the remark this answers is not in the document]
jamesturner246: Agree. [Some background](https://kafka.apache.org/20/javadoc/org/apache/kafka/clients/producer/ProducerRecord.html), saying it's producer-side configured, so need to ask.

main/management/commands/kafka_consumer.py
```diff
@@ -31,21 +32,46 @@ def handle(self, debug: bool = False, **kwargs: Any) -> None:  # type: ignore[mi
...
+                    # atomic here to prevent race condition with messages being
+                    # displayed by the web application
+                    query = DruncMessage.objects.filter(timestamp__lt=expire_time)
+                    if query.count():
+                        if debug:
+                            self.stdout.write(
+                                f"Deleting {query.count()} messages "
+                                f"older than {expire_time}."
+                            )
+                        query.delete()
```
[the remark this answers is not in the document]
jamesturner246: Oops.. :+1:

process_manager/templates/process_manager/index.html
dalonsoa, line 1: I thought this had been done in a different PR, so I'm not sure why it is here again.

process_manager/templates/process_manager/partials/message_items.html
dalonsoa, line 1: Same here.

process_manager/views/partials.py
```diff
@@ -62,13 +61,8 @@ def process_table(request: HttpRequest) -> HttpResponse:
...
 def messages(request: HttpRequest) -> HttpResponse:
-    """Renders and pops Kafka messages from the user's session."""
-    with transaction.atomic():
-        # atomic to avoid race condition with kafka consumer
-        messages = request.session.load().get("messages", [])
-        request.session.pop("messages", [])
-        request.session.save()
-
+    """Renders Kafka messages from the database."""
+    messages = [f"{msg.timestamp}: {msg.message}" for msg in DruncMessage.objects.all()]
```
dalonsoa, line 65: In theory, we should be pulling the messages only relevant for this user, which means that there should be user-related information in the `DruncMessage` model, right?

dalonsoa (changes requested):
It looks ok, but I think we have lost the user-specificity of the messages we had before. For account for that, the `DruncMessage` model will need to have an owner or a user or something like that, and add the messages from Kafka with that user information.

process_manager/views/partials.py
cc-a: That was my initial assumption but after checking with the project team all users need to see all messages so there isn't a requirement for this.

process_manager/templates/process_manager/index.html
jamesturner246: I remember it being done for the process table

process_manager/views/partials.py
dalonsoa: Odd, but so be it.

jamesturner246:
Done. Managed to do the timezone stuff with django-only fns. @dalonsoa, are your comments addressed?

dalonsoa (approved):
QA checks are failing, but otherwise the code looks good!

main/management/commands/kafka_consumer.py
```diff
@@ -31,21 +31,37 @@ def handle(self, debug: bool = False, **kwargs: Any) -> None:  # type: ignore[mi
...
             for messages in consumer.poll(timeout_ms=500).values():
+                message_timestamps = []
                 message_bodies = []
                 for message in messages:
                     if debug:
                         self.stdout.write(f"Message received: {message}")
                         self.stdout.flush()
+
+                    # Convert Kafka timestamp (milliseconds) to datetime (seconds).
+                    timestamp = datetime.fromtimestamp(message.timestamp / 1e3, tz=UTC)
```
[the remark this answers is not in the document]
jamesturner246: :thinking: 

```
<ipython-input-11-d286e0fff8ad>:1: DeprecationWarning: datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.fromtimestamp(timestamp, datetime.UTC).
```

Trend is towards overdefinedness.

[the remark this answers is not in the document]
cc-a: Ah, fair enough.

process_manager/views/partials.py
```diff
@@ -62,12 +62,11 @@ def process_table(request: HttpRequest) -> HttpResponse:
...
-    """Renders and pops Kafka messages from the user's session."""
-    with transaction.atomic():
-        # atomic to avoid race condition with kafka consumer
-        messages = request.session.load().get("messages", [])
-        request.session.pop("messages", [])
-        request.session.save()
+    """Renders Kafka messages from the database."""
+    messages = []
+    for msg in DruncMessage.objects.all():
+        timestamp = localtime(msg.timestamp).strftime("%Y-%m-%d %H:%M:%S")
```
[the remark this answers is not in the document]
jamesturner246: :+1:  Done.

jamesturner246:
Sorry, should have asked first: do you prefer to do the merging in future @cc-a ?

PR #147, opened 2024-10-10
Use service profiles to specify drunc-lite use case

### LINKED ISSUE #115: Docker image for controller development
The core functionality of this has been addressed in #81 but due to the large image size, it would be good to take a pass at this with a view on the developer experience i.e. it would be nice to be able to work with only the smaller process manager where needed. Possibly we could use [profiles](https://docs.docker.com/compose/how-tos/profiles/) for this?

TinyMarsh:
# Description

Uses profiles in Compose to allow the user to selectively activate either the full `drunc` service, or the `dunc-lite` service.

To test run either;
```sh
docker compose --profile drunc up
```
or
```sh
docker compose --profile drunc-lite up
```
for the full or lite services, respectively.

Fixes #115

## Type of change

## Key checklist

## Further checks


cc-a (commented):
Looks promising but its based on a fairly old version of main at this point and a couple of other services have been added to docker-compose.yml. Could you update?

alexdewar (approved):
LGTM except for the `pre-commit` thing!

I'm guessing the lite image is exactly the same as we had before...?

TinyMarsh:
Some thoughts on continuing the work in this PR...

After bringing this branch up to date with `main`, I noticed that we would need to find a way to dynamically pass the correct value to the `PROCESS_MANAGER_URL` environment variable;
https://github.com/ImperialCollegeLondon/drunc_ui/blob/d09c1890272b6ee0e6ee478a39f6820004e23957/docker-compose.yml#L16

I was considering handling this with a bash script, where the user can pass by argument which drunc service to run and, at the same time, set the `PROCESS_MANAGER_URL` by setting a Bash environment variable (`DRUNC_SERVICE`). For example:
```bash
#scripts/run_docker_compose.sh
#!/bin/bash

PROFILE=$1

# Set environment variable based on the profile
if [ "$PROFILE" == "drunc-lite" ]; then
  export DRUNC_SERVICE=drunc-lite
else
  export DRUNC_SERVICE=drunc
fi

# Run Docker Compose with the specified profile
docker compose --profile $PROFILE up
```

Then, in `docker-compose.yml`:
```YAML
services:
  app:
    build: .
    command:
      - bash
      - -c
      - |
        python manage.py migrate
        python manage.py runserver 0:8000
    ports:
      - 127.0.0.1:8000:8000
    volumes:
      - .:/usr/src/app
      - db:/usr/src/app/db
    environment:
      - PROCESS_MANAGER_URL=${DRUNC_SERVICE}:10054
    depends_on:
      - ${DRUNC_SERVICE}
```

It feels a bit hacky, but it could work. From the user perspective it's still just running one line; `./scripts/run_docker_compose.sh drunc-lite`. Unfortunately, whilst implementing this, the app seems to have stopped working on my system (even when I have `main` checked out 😞) and I ran out of time debugging that.

PR #148, opened 2024-10-11
Use hx-preserve with table checkboxes

cc-a:
# Description

This PR exempts the table checkboxes from being updated by HTMX requests. The main benefits:
- removes issues with flickering checkboxes
- allows removal of all server side handling of checkbox state

It does add the need to handle the header checkbox state to the client side but this is handled with a simple hyperscript snippet. For consistency I've also replaced the vanilla javascript with hyperscript to set row checkboxes when the header is clicked.

## Type of change

- [X] Optimization (non-breaking, back-end change that speeds up the code)
- [X] Bug fix (non-breaking change which fixes an issue)

## Key checklist

## Further checks

- [X] Code is commented, particularly in hard-to-understand areas
- [X] Tests added or an issue has been opened to tackle that in the future. (Indicate issue here: # (issue))


process_manager/tables.py
```diff
@@ -43,6 +54,23 @@ class Meta:  # noqa: D106
...
+        attrs={
+            "th__input": {
+                "id": "header-checkbox",
+                "hx-preserve": "true",
+                "_": header_checkbox_hyperscript,
+            }
+        },
     )
+
+    def render_select(self, value: str) -> str:
```
dalonsoa, line 66: What's `value` here? Is the `uuid` of the column? Whatever it is, maybe add it to the docstring.

Also, I'm not sure when this method is called, or by whom.

process_manager/tables.py
```diff
@@ -43,6 +54,23 @@ class Meta:  # noqa: D106
     select = tables.CheckBoxColumn(
         accessor="uuid",
         verbose_name="Select",
-        checked="checked",
-        attrs={"th__input": {"onclick": "toggle(this)"}},
+        attrs={
+            "th__input": {
+                "id": "header-checkbox",
+                "hx-preserve": "true",
+                "_": header_checkbox_hyperscript,
```
dalonsoa, line 61: What does this underscore represent?

process_manager/templates/process_manager/partials/process_table.html
```diff
@@ -1,4 +1,4 @@
 {% load render_table from django_tables2 %}
-<div hx-post="{% url 'process_manager:process_table' %}"
-     hx-trigger="load delay:1s, click from:input"
+<div hx-get="{% url 'process_manager:process_table' %}"
```
dalonsoa, line 2: Is it `get` now because we are no longer sending data (the selected rows) to the backend?

process_manager/tables.py
```diff
@@ -26,6 +27,16 @@
     "<a href=\"{% url 'process_manager:logs' record.uuid %}\">LOGS</a>"
 )
 
+header_checkbox_hyperscript = "on click set .row-checkbox.checked to my.checked"
+
+row_checkbox_hyperscript = """
+on click
+if <input.row-checkbox:not(:checked)/> is empty
+  set #header-checkbox.checked to true
```
dalonsoa, line 35: In plain(er) English, this means that if the list of rows that are not checked is empty, then the header checkbox should be checked - as all the rows are checked - right?

dalonsoa (approved):
It looks ok and it works, so approving. Having said that, I've lots of questions.

process_manager/tables.py
cc-a: Correct.

process_manager/tables.py
cc-a: That is the html attribute that hyperscript uses to store its scripts on individual elements.

process_manager/templates/process_manager/partials/process_table.html
cc-a: Right.

process_manager/tables.py
cc-a: Thanks good catch. I wish there was a way to get automatic checks for this but I've never been able to find one.

It's called during table rendering. It's detected by class magicness and called in place of the render method of the column.

cc-a:
This has been up for a little while now so I'll merge by the end of the day unless anyone else has any feedback/questions.

process_manager/tables.py
alexdewar: `mkdocs` gives warnings about malformed docstrings (=errors if you pass `--strict`), but I don't know if it warns about missing args. Were we going to add API documentation at some point?

tests/process_manager/views/test_partial_views.py
```diff
@@ -16,13 +15,16 @@ class TestProcessTableView(LoginRequiredTest):
...
-        response = getattr(auth_client, method)(self.endpoint)
+        uuids = [str(uuid4()) for _ in range(5)]
+        self._mock_session_info(mocker, uuids)
+        response = auth_client.get(self.endpoint)
         assert response.status_code == HTTPStatus.OK
-        assert isinstance(response.context["table"], ProcessTable)
+        table = response.context["table"]
+        assert isinstance(table, ProcessTable)
+        for row, uuid in zip(table.data.data, uuids):
+            assert row["uuid"] == uuid
```
alexdewar, line 27:
```suggestion
        assert all(row["uuid"] == uuid for row, uuid in zip(table.data.data, uuids))
```

alexdewar (approved):
LGTM insofar as I understand it :wink:

process_manager/tables.py
```diff
@@ -43,6 +54,28 @@ class Meta:  # noqa: D106
...
+        Overriding the default render method for this column is required as the
+        hx-preserve attitribute requires all elements to have unique id values. We also
+        need to add the hyperscript required for the header checkbox behaviour.
+
+        Called during table rendering.
+
+        Args:
+          value: uuid from the table row data
+        """
+        return mark_safe(
```
jamesturner246, line 78: So `render_select` is magically called when rendering `self.select = tables.CheckBoxCol...`, and creates a custom `id` for the row checkbox, so that `hx-preserve` can preserve each row cbox on refresh.

What is `mark_safe` doing here exactly?

process_manager/tables.py
```diff
@@ -26,6 +27,16 @@
     "<a href=\"{% url 'process_manager:logs' record.uuid %}\">LOGS</a>"
 )
 
+header_checkbox_hyperscript = "on click set .row-checkbox.checked to my.checked"
+
+row_checkbox_hyperscript = """
+on click
+if <input.row-checkbox:not(:checked)/> is empty
```
jamesturner246, line 34: Why `input.row-checkbox` and not `.row-checkbox` like above?

jamesturner246 (approved):
Looks good. Just questions.

process_manager/tables.py
cc-a: Planning to add it for next sprint.

process_manager/tables.py
cc-a: https://docs.djangoproject.com/en/5.1/ref/utils/#django.utils.safestring.mark_safe It tells the Django templating system that the html generated is safe and isn't vurnerable to injection attacks so it can just be rendered directly. Otherwise the templating system by default will try to escape everything.

process_manager/tables.py
cc-a: No good reason. I'll make them consistent.

PR #158, opened 2024-10-18
Kafka message filtering

### LINKED ISSUE #122: Filter message feed
Because of the anticipated large number of messages it's important to be able to filter messages based on text contents. We can pattern this off of - https://htmx.org/examples/active-search/.

- Add a partial template for the message feed and include it in the main feed template.
- Add a form with an input field at the top of the message feed (use `method="get"` on the form to send data as query parameters).
- Add a view function that takes a filter query parameter and returns the rendered message feed with the filter provided.
- Add htmx magic to the input field. Something like: `hx-get="/messages_endpoint/" hx-target="#message-feed" hx-swap="outerHTML"`.

If #121 has been implemented then filtering should also be added to the view function that adds new messages to the feed.

jamesturner246:
# Description

This PR enables filtering of Kafka messages in the message feed. Searching is case-insensitive. The timestamp is *not* considered by the search.

Screenshot from 2024-10-18 14-02-24

Also tests are added, and partial template/triggers are cleaned a little.

Fixes #122 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/process_manager/templates/process_manager/index.html
+++ b/process_manager/templates/process_manager/index.html
@@ -53,14 +53,12 @@
                type="search"
                name="search"
                placeholder="Search..."
-               hx-get="{% url 'process_manager:process_table' %}"
                hx-trigger="input changed delay:500ms, every 1s"
-               hx-target="div.table-container"
-               hx-swap="innerHTML">
+               hx-get="{% url 'process_manager:process_table' %}"
+               hx-target="div.table-container">
         <div id="table-container"
-             hx-get="{% url 'process_manager:process_table' %}"
-             hx-swap="innerHTML"
-             hx-trigger="load"></div>
+             hx-trigger="load"
+             hx-get="{% url 'process_manager:process_table' %}"></div>
       </form>
     </div>
     <div class="col" id="message-panel">
@@ -73,13 +71,10 @@
                   aria-label="Hide Messages"
                   _="on click hide #message-panel show #show-messages-button"></button>
         </div>
-        <div class="card-body" id="message-list">
-          <ul class="list-group">
-            <div hx-get="{% url 'process_manager:messages' %}"
-                 hx-swap="outerHTML"
-                 hx-trigger="load"></div>
-          </ul>
-        </div>
+        <div id="message-list"
+             class="card-body"
+             hx-trigger="load, every 1s"
+             hx-get="{% url 'process_manager:messages' %}"></div>
       </div>
     </div>
   </div>
--- a/process_manager/templates/process_manager/index.html
+++ b/process_manager/templates/process_manager/index.html
@@ -53,12 +53,12 @@
                type="search"
                name="search"
                placeholder="Search..."
-               hx-trigger="input changed delay:500ms, every 1s"
                hx-get="{% url 'process_manager:process_table' %}"
-               hx-target="#table-container">
+               hx-trigger="input changed delay:500ms, every 1s"
+               hx-target="div.table-container">
         <div id="table-container"
-             hx-trigger="load"
-             hx-get="{% url 'process_manager:process_table' %}"></div>
+             hx-get="{% url 'process_manager:process_table' %}"
+             hx-trigger="load"></div>
       </form>
     </div>
     <div class="col" id="message-panel">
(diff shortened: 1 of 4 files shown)


process_manager/templates/process_manager/index.html
```diff
@@ -53,14 +53,12 @@
...
-               hx-target="div.table-container"
-               hx-swap="innerHTML">
+               hx-get="{% url 'process_manager:process_table' %}"
+               hx-target="#table-container">
         <div id="table-container"
-             hx-get="{% url 'process_manager:process_table' %}"
-             hx-swap="innerHTML"
-             hx-trigger="load"></div>
+             hx-trigger="load"
+             hx-get="{% url 'process_manager:process_table' %}"></div>
```
cc-a, line 61: What are these changes for? Diff is a bit confusing as some lines are being moved around. Could you tidy it a bit to minimise the diff?

process_manager/views/partials.py
```diff
@@ -96,10 +96,17 @@ def process_table(request: HttpRequest) -> HttpResponse:
...
 def messages(request: HttpRequest) -> HttpResponse:
     """Renders Kafka messages from the database."""
+    search = request.GET.get("search", "").lower()
+
     messages = []
     for msg in DruncMessage.objects.all():
+        # Filter messages based on search parameter.
+        if search not in msg.message.lower():
+            continue
+
```
cc-a, line 106: Can make this simpler and more efficient with `DruncMessage.objects.filter(message__icontains=search)`.

tests/process_manager/views/test_partial_views.py
```diff
@@ -83,6 +83,41 @@ def test_get(self, auth_client):
         t2_str = t2.strftime("%Y-%m-%d %H:%M:%S")
         assert response.context["messages"][0] == f"{t2_str}: message 1"
 
+    def test_get_with_search(self, auth_client):
+        """Tests message filtering of view method."""
+        from datetime import UTC, datetime
+
+        from main.models import DruncMessage
```
cc-a, line 90: These imports are repeated in the test above so lets move them to the global scope.

cc-a (changes requested):
Good stuff. Just a couple of small tweaks.

process_manager/templates/process_manager/index.html
jamesturner246: Diff was a bit messy, but just removed `innerHTML` as it's the default anyway. Fixed the diff.

cc-a (approved):
Lovely. I'll leave this up for a bit before merging in case someone else wants to take a pass.

process_manager/views/partials.py
```diff
@@ -97,9 +97,13 @@ def process_table(request: HttpRequest) -> HttpResponse:
 def messages(request: HttpRequest) -> HttpResponse:
     """Renders Kafka messages from the database."""
     messages = []
-    for msg in DruncMessage.objects.all():
+
+    # Filter messages based on search parameter.
+    search = request.GET.get("search", "")
+    for msg in DruncMessage.objects.filter(message__icontains=search):
         # Time is stored as UTC. localtime(t) converts this to our configured timezone.
         timestamp = localtime(msg.timestamp).strftime("%Y-%m-%d %H:%M:%S")
```
dalonsoa, line 105: Not related to this PR, but `localtime` returns the actual local time _if such a time is set somewhere explicitl_y. Otherwise, Django will just return the default timezone - typically UTC. [Django does not autodetect the user's timezone](https://docs.djangoproject.com/en/5.1/topics/i18n/timezones/#selecting-the-current-time-zone), as far as I learnt lately.

dalonsoa (approved):
LGTM!

process_manager/views/partials.py
jamesturner246: Oh strange. I thought it picked up the one defined in `drunc_ui/settings/settings.py`?

jamesturner246: As in [here](https://docs.djangoproject.com/en/5.1/topics/i18n/timezones/#time-zone-aware-output-in-templates), also `USE_TZ` and `TIME_ZONE` [here](https://docs.djangoproject.com/en/5.1/topics/i18n/timezones/#default-time-zone-and-current-time-zone).

dalonsoa: Precisely, and in there is says:

> You should set the current time zone to the end user’s actual time zone with [activate()](https://docs.djangoproject.com/en/5.1/ref/utils/#django.utils.timezone.activate). Otherwise, the default time zone is used.

But Django cannot get the user's actual timezone automatically, as explained exacly in the next section, that I linked above.

PR #176, opened 2024-10-22
Adding bootswatch theme

### LINKED ISSUE #72: Use a Bootswatch theme
The in-built Bootstrap theme isn't the best colour scheme. [Bootswatch](https://bootswatch.com) have many free modern web themes that we can use.

See [QwikApp settings](https://github.com/ImperialCollegeLondon/QwikApp_web/blob/main/qwikapp/settings/settings.py#L208C1-L221C2) for how to Include them

Sahil590:
# Description

This PR adds the [Bootswatch](https://bootswatch.com) theme, specifically Sandstone (this can be changed to comment if you prefer any others, just going off what was discussed in the issue)
In addition, added the bootstrap JS file to the static folder, added the script tag to base.html

Fixes #72  (issue)

## Type of change

## Key checklist

## Further checks


cc-a (approved):
Nice. I like it. As discussed the javascript is also being served as a static file as the production infrastructure is not expected to have access to the public internet so the CDN based approach we had previously won't work.

main/templates/main/base.html
```diff
@@ -11,17 +10,17 @@
...
     {% endblock extra_css %}
     {% block extra_js %}
     {% endblock extra_js %}
     <script src="{% static 'main/js/htmx.min.js' %}"></script>
     <script src="{% static 'main/js/_hyperscript.min.js' %}"></script>
+    <script src="{% static 'main/js/bootstrap.bundle.min.js' %}"></script>
     <link rel="shortcut icon"
           type="image/png"
           href="{% static 'main/images/Favicon.png' %}">
+    <link rel="stylesheet" href="{% static 'main/css/bootstrap.min.css' %}">
```
dalonsoa, line 23: So, the theme is set by using this CSS file, not by setting any flag or configuration setting to `sandstone` anywhere?

Sahil590: Yeah on the bootswatch website you can get the css for each theme

dalonsoa (approved):

PR #179, opened 2024-10-24
Disable action buttons if not checkbox is selcted

### LINKED ISSUE #90: Hide action buttons based on permissions
Requires #12 and #86.

Users without the permissions required to access the restart/kill/flush view functions should not be shown the buttons in the UI. Need some template logic to control their rendering.

See #85 for reference.

### LINKED ISSUE #110: Deactivate restart/flush/kill if no processes are selected
If no rows are checked in the process table then it would be nice to disable the buttons that carry out actions based on these.

Sahil590:
# Description
adds some logic to disable the restart, flush and kill if not checkboxes are selected. 
I have also added the permission to be able to access the button to the permission group "can_modify_processes"

Fixes #90 and #110 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/process_manager/tables.py
+++ b/process_manager/tables.py
@@ -57,5 +57,6 @@ class ProcessTable(tables.Table):
         """
         return mark_safe(
             f'<input type="checkbox" name="select" value="{value}" id="{value}-input" '
-            f'hx-preserve="true" class="row-checkbox" _="{row_checkbox_hyperscript}"'
+            f'hx-preserve="true" class="row-checkbox" _="{row_checkbox_hyperscript}" '
+            f'onchange="updateButtonState()">'
         )
--- a/process_manager/tables.py
+++ b/process_manager/tables.py
@@ -9,14 +9,6 @@ logs_column_template = (
 
 header_checkbox_hyperscript = """
 on click set .row-checkbox.checked to my.checked
-if my.checked
-    set #killButton.disabled to false
-    set #restartButton.disabled to false
-    set #flushButton.disabled to false
-else
-    set #killButton.disabled to true
-    set #restartButton.disabled to true
-    set #flushButton.disabled to true
 """
 
 row_checkbox_hyperscript = """
@@ -67,5 +59,5 @@ class ProcessTable(tables.Table):
         """
         return mark_safe(
             f'<input type="checkbox" name="select" value="{value}" id="{value}-input" '
-            f'hx-preserve="true" class="row-checkbox" _="{row_checkbox_hyperscript}" '
+            f'hx-preserve="true" class="row-checkbox" _="{row_checkbox_hyperscript}"'
         )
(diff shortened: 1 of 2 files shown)


cc-a (changes requested):
Hi @Sahil590. As noted in a comment on the issue, for consistency with our other front ending scripting it would be preferable for this to be written in [hyperscript](https://hyperscript.org/). It should be possible to do it as a single line for each button that way. Let me know if you'd like to have a call to discuss.

process_manager/tables.py
```diff
@@ -7,7 +7,17 @@
...
+header_checkbox_hyperscript = """
+on click set .row-checkbox.checked to my.checked
+if my.checked
+    set #killButton.disabled to false
+    set #restartButton.disabled to false
+    set #flushButton.disabled to false
+else
+    set #killButton.disabled to true
+    set #restartButton.disabled to true
+    set #flushButton.disabled to true
```
cc-a, line 19: This has an unexpected behaviour - if there are no rows in the table, toggling the header checkbox still changes the state of the buttons. The state of the buttons should only be determined by whether there are any checked row checkboxes not by the state of the header checkbox. Can the `disableActionButton` behaviour just be extended to fire when the header checkbox is clicked (in addition to the row-checkboxes)?

process_manager/tables.py
```diff
@@ -57,5 +67,5 @@ def render_select(self, value: str) -> str:
         """
         return mark_safe(
             f'<input type="checkbox" name="select" value="{value}" id="{value}-input" '
-            f'hx-preserve="true" class="row-checkbox" _="{row_checkbox_hyperscript}"'
+            f'hx-preserve="true" class="row-checkbox" _="{row_checkbox_hyperscript}" '
```
cc-a, line 70: Removable.

process_manager/templates/process_manager/index.html
```diff
@@ -14,12 +14,15 @@
...
-  }
+  <script type="text/hyperscript">
+  behavior disableActionButton
+    on click[target.matches('.row-checkbox')] from elsewhere
+      if <.row-checkbox:checked /> is empty
+        set me.disabled to true
+      else
+        set me.disabled to false
+    end
+    end
```
cc-a, line 25: Indentation.

process_manager/templates/process_manager/index.html
```diff
@@ -29,21 +32,32 @@
...
-               value="Flush"
-               class="btn btn-warning"
-               name="action"
-               onclick="return confirm('Flush selected processes?')">
-        <input type="submit"
-               value="Kill"
-               class="btn btn-danger"
-               name="action"
-               onclick="return confirm('Kill selected processes?')">
+        {% if perms.main.can_modify_processes %}
```
cc-a, line 35: Should be applied to the boot button as well.

cc-a (changes requested):
Looking good. Couple of tweaks to look at.

process_manager/tables.py
Sahil590: I tried to add it and it works, but it has some strange behaviour

dalonsoa (approved):
I have not been able to test it due to #187, but the code looks ok.

cc-a (approved):
Good stuff.

PR #188, opened 2024-10-29
Neater UI

cvzbynek:
# Description

This PR improves the user interface for the process manager page, introducing several enhancements for a more polished, functional layout:
1. Adjusted the **message panel** to maximize screen space usage.
2. Implemented **Flexbox layout** for the main container to allow the process control table to expand to full width when needed.
3. Minor CSS adjustments to improve table and button spacing, uniform button appearance, and column alignment.
4. **Logs** screen revised

Fixes #170 

## Type of change

- [ ] Documentation (non-breaking change that adds or improves the documentation)
- [x] New feature (non-breaking change which adds functionality)
- [x] Optimization (non-breaking, back-end change that speeds up the code)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (whatever its nature)

## Key checklist

- [ ] All tests pass (eg. `python -m pytest`)
- [ ] The documentation builds and looks OK (eg. `python -m sphinx -b html docs docs/build`)
- [ ] Pre-commit hooks run successfully (eg. `pre-commit run --all-files`)

## Further checks

- [x] Code is commented, particularly in hard-to-understand areas
- [ ] Tests added or an issue has been opened to tackle that in the future.

PR #192, opened 2024-10-30
Adds Time info to message feed header

### LINKED ISSUE #159: Add timezone to message feed header
Kafka messages are stored in the database with UTC timestamps. When rendered in the message feed these are converted into the localtime based on the django `TIME_ZONE` setting. At the moment there is no indication of what timezone is being used. We should add this info into the header of the message feed - preferably the geographic timezone as well as the UTC offset (e.g. "London/Europe (UTC+1)".

Sahil590:
# Description
uses JavaScript to get the timezone info and displays if in the message feed header. 

Fixes #159  (issue)

## Type of change

## Key checklist

## Further checks


process_manager/templates/process_manager/index.html
```diff
@@ -64,6 +64,17 @@
...
   </script>
+  <script>
+    document.addEventListener('DOMContentLoaded', function() {
+      var timezoneElement = document.getElementById('timezone');
+      var timezone = Intl.DateTimeFormat().resolvedOptions().timeZone;
+      var offset = new Date().getTimezoneOffset();
+      var offsetHours = Math.abs(offset / 60);
+      var offsetSign = offset <= 0 ? '+' : '-';
+      var offsetString = offsetSign + String(offsetHours).padStart(2, '0')
+      timezoneElement.textContent = timezone + ' (UTC' + offsetString + ')';
```
dalonsoa, line 75: My JS knowledge is non-existent, so just to confirm I get this right, this line here will update the HTML element below with `id="timezone"` giving it the value `TIMEZONE (UTC + OFFSET_HOURS)`, both TIMEZONE and OFFSET calculated here, right?

I'd add `h` to the string. Not sure if standard, but makes it clearer. 
```suggestion
      timezoneElement.textContent = timezone + ' (UTC' + offsetString + 'h)';
```

Also, there are some [fractional timezones](https://www.timeanddate.com/time/time-zones-interesting.html) which I'm not sure if will be handled properly with this.

dalonsoa (approved):
LGTM. Just a clarification for my own benefit.

process_manager/templates/process_manager/index.html
Sahil590: Yeah, exactly it will update the element with that ID. 
I did not take into account the fractional timezones will amend this.

Sahil590: Currently, it shows this if it is a fractional timezone
image

cc-a (changes requested):
Unfortunately this needs to use a different approach. Message timestamps are rendered according to the django `TIME_ZONE` setting i.e. a fixed timezore for all users. This implementation will dynamically detect the users current timezone rather than the timezone configured on the server. As this setting is only available on the server the rendering should be performed an the server side within the template.

Sahil590:
I followed these docs to get the server time https://docs.djangoproject.com/en/5.1/ref/templates/builtins/#std:templatefilter-time
but I get it in this format without the region @cc-a does this work? 
image

process_manager/templates/process_manager/index.html
```diff
@@ -126,28 +127,32 @@ <h5>Process Control</h5>
...
-            <input class="form-control mb-2"
-                   type="search"
-                   name="search"
-                   placeholder="Search messages..."
-                   hx-get="{% url 'process_manager:messages' %}"
-                   hx-trigger="input changed delay:500ms, every 1s"
-                   hx-target="#message-list">
-            <div id="message-list"
-                 class="list-group"
+            {% timezone None %}
```
cc-a, line 130: We're not using timezone localisation so this tag shouldn't be necessary.

cc-a (commented):
Great. Needs a little visual tweak I think. Maybe it could be shown as a small grey line of text above the search bar - "All timestamps are displayed in ...".

Sahil590:
@cc-a should be fine now tests are failing due to #188

PR #195, opened 2024-10-31
adding decorator to messages view

### LINKED ISSUE #150: Make htmx polling more robust
We're using htmx polling to update the message feed and process table. In the event of a 500 response stops polling. This has been observed in cases where gRPC calls to the process manager fail and results in the process table no longer updating. We should prevent partial view functions from returning 500 codes and instead return content with a useful error messoge.

Not totally sure this will work but I suggest we create our own View subclass that:
- has a configurable attribute for a custom error message.
- wraps `View.dispatch` with an `except Exception` block which renders a standard `partials/error.html` template containing the custom error message.
- logs the underlying exception.

Then we write all our partial view functions using that class.

Sahil590:
# Description

I have added an error handling decorator for the message feed. 

Fixes #150 (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/process_manager/views/partials.py
+++ b/process_manager/views/partials.py
@@ -17,6 +17,33 @@ from ..process_manager_interface import get_session_info
 from ..tables import ProcessTable
 
 
+def handle_errors(
+    view_func: Callable[[HttpRequest], HttpResponse],
+) -> Callable[[HttpRequest], HttpResponse]:
+    """Decorator to handle errors.
+
+    Args:
+        view_func: The view function to be wrapped.
+
+    Returns:
+        The wrapped view function.
+    """
+    logger = logging.getLogger("django")
+
+    def wrapped_view(request, *args, **kwargs) -> HttpResponse:  # type: ignore
+        try:
+            return view_func(request, *args, **kwargs)
+        except Exception as e:
+            logger.exception(e)
+            return render(
+                request,
+                "process_manager/partials/error_message.html",
+                {"error_message": "An error occurred while processing your request."},
+            )
+
+    return wrapped_view
+
+
 def filter_table(
     search: str, table: list[dict[str, str | int]]
 ) -> list[dict[str, str | int]]:
@@ -56,6 +83,7 @@ def filter_table(
 
 
 @login_required
+@handle_errors
 def process_table(request: HttpRequest) -> HttpResponse:
     """Renders the process table.
 
@@ -95,33 +123,6 @@ def process_table(request: HttpRequest) -> HttpResponse:
         context=dict(table=table),
         template_name="process_manager/partials/process_table.html",
     )
-
-
-def handle_errors(
-    view_func: Callable[[HttpRequest], HttpResponse],
-) -> Callable[[HttpRequest], HttpResponse]:
-    """Decorator to handle errors.
-
-    Args:
-        view_func: The view function to be wrapped.
-
-    Returns:
-        The wrapped view function.
-    """
-    logger = logging.getLogger("django")
-
-    def wrapped_view(request, *args, **kwargs) -> HttpResponse:  # type: ignore
-        try:
-            return view_func(request, *args, **kwargs)
-        except Exception as e:
-            logger.exception(f"Error in view {view_func.__name__}: {e}")
-            return render(
-                request,
-                "process_manager/partials/error_message.html",
-                {"error_message": "An error occurred while processing your request."},
-            )
-
-    return wrapped_view
 
 
 @login_required
--- a/process_manager/views/partials.py
+++ b/process_manager/views/partials.py
@@ -35,11 +35,7 @@ def handle_errors(
             return view_func(request, *args, **kwargs)
         except Exception as e:
             logger.exception(e)
-            return render(
-                request,
-                "process_manager/partials/error_message.html",
-                {"error_message": "An error occurred while processing your request."},
-            )
+            return render(request, "process_manager/partials/error_message.html")
 
     return wrapped_view
 
--- a/process_manager/views/partials.py
+++ b/process_manager/views/partials.py
@@ -108,13 +108,13 @@ def handle_errors(
     Returns:
         The wrapped view function.
     """
-    logger = logging.getLogger(__name__)
+    logger = logging.getLogger("django")
 
     def wrapped_view(request, *args, **kwargs) -> HttpResponse:  # type: ignore
         try:
             return view_func(request, *args, **kwargs)
         except Exception as e:
-            logger.error(f"Error in view {view_func.__name__}: {e}")
+            logger.exception(f"Error in view {view_func.__name__}: {e}")
             return render(
                 request,
                 "process_manager/partials/error_message.html",
--- a/process_manager/views/partials.py
+++ b/process_manager/views/partials.py
@@ -1,5 +1,8 @@
 """View functions for partials."""
 
+import logging
+from collections.abc import Callable
+
 import django_tables2
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


Sahil590:
@cc-a It's still in draft, but is this along the lines of what we want to achieve?

process_manager/views/partials.py
```diff
@@ -94,7 +97,35 @@ def process_table(request: HttpRequest) -> HttpResponse:
...
+) -> Callable[[HttpRequest], HttpResponse]:
+    """Decorator to handle errors.
+
+    Args:
+        view_func: The view function to be wrapped.
+
+    Returns:
+        The wrapped view function.
+    """
+    logger = logging.getLogger(__name__)
```
cc-a, line 111: Probably want the `django` logger here.

process_manager/views/partials.py
```diff
@@ -94,7 +97,35 @@ def process_table(request: HttpRequest) -> HttpResponse:
...
+
+    def wrapped_view(request, *args, **kwargs) -> HttpResponse:  # type: ignore
+        try:
+            return view_func(request, *args, **kwargs)
+        except Exception as e:
+            logger.error(f"Error in view {view_func.__name__}: {e}")
```
cc-a, line 117: Can use [logger.exception](https://docs.python.org/3/library/logging.html#logging.Logger.exception) here.

cc-a (commented):
Yeah, that's pretty much it.

process_manager/views/partials.py
```diff
@@ -94,7 +97,35 @@ def process_table(request: HttpRequest) -> HttpResponse:
...
+    logger = logging.getLogger("django")
+
+    def wrapped_view(request, *args, **kwargs) -> HttpResponse:  # type: ignore
+        try:
+            return view_func(request, *args, **kwargs)
+        except Exception as e:
+            logger.exception(f"Error in view {view_func.__name__}: {e}")
+            return render(
+                request,
+                "process_manager/partials/error_message.html",
```
cc-a, line 120: We're missing this file.

process_manager/views/partials.py
```diff
@@ -94,7 +97,35 @@ def process_table(request: HttpRequest) -> HttpResponse:
...
+    Returns:
+        The wrapped view function.
+    """
+    logger = logging.getLogger("django")
+
+    def wrapped_view(request, *args, **kwargs) -> HttpResponse:  # type: ignore
+        try:
+            return view_func(request, *args, **kwargs)
+        except Exception as e:
+            logger.exception(f"Error in view {view_func.__name__}: {e}")
```
cc-a, line 117: `logger.exception` will take care of recording all of the exception details so you don't need to include it in the error message.

process_manager/views/partials.py
```diff
@@ -94,7 +97,35 @@ def process_table(request: HttpRequest) -> HttpResponse:
...
+                request,
+                "process_manager/partials/error_message.html",
+                {"error_message": "An error occurred while processing your request."},
+            )
+
+    return wrapped_view
+
+
 @login_required
+@handle_errors
```
cc-a, line 128: Also need to apply the decorator to the other view function in the file.

cc-a (changes requested):
The implementation looks good. Couple of small comments and I think we're missing the error template.

It would be good to have tests for the process_table and messages view that checks that this behaviour has been applied.

process_manager/views/partials.py
Sahil590: Was going to ask what should i write for this

cc-a: Probably just "error refresh page content" in red text will do.

Sahil590: Would I have to include the same template tags as the message_items.html

cc-a: No, just a span or something with the text in.

process_manager/views/partials.py
```diff
@@ -14,6 +17,33 @@
...
+
+    def wrapped_view(request, *args, **kwargs) -> HttpResponse:  # type: ignore
+        try:
+            return view_func(request, *args, **kwargs)
+        except Exception as e:
+            logger.exception(e)
+            return render(
+                request,
+                "process_manager/partials/error_message.html",
+                {"error_message": "An error occurred while processing your request."},
```
cc-a, line 41: The template doesn't render the `error_message` variable so this can be removed.

cc-a (approved):
Great. One extra line you can take out.

process_manager/views/partials.py
dalonsoa: So we are not actually providing any useful error message to the user, just that something went wrong, right?

dalonsoa (approved):
Code looks good to me!

process_manager/views/partials.py
cc-a: Yeah. With the current implementation if even one request fails then updates to the table/message feed stop but there is no sign of this to the user. This is an improvement in that in the case of an error there will be some indication to the user but if it was a one-off then the page will update again removing the error if not they have the chance to refresh or investigate further.

jamesturner246 (commented):
Okay, so if an error occurs in, say, `messages`, then only the substituted message pane contents will display error, but the  rest of the page continues to operate (and periodically retrying the partials)?

Is it worth adding a test which mocks an exception-throwing partial to check if the error partial loads?

Sahil590:
@jamesturner246 yes if an error occurs in the message and not in the process table, it will still get the partials. 
I forgot about the tests and will add them shortly.

jamesturner246 (approved):
LGTM! :+1:

tests/process_manager/views/test_partial_views.py
```diff
@@ -193,3 +196,32 @@ def test_filter_table(search, table, expected):
     from process_manager.views.partials import filter_table
 
     assert filter_table(search, table) == expected
+
+
+class HandleErrorsTest(TestCase):
```
AdrianDAlessandro, line 201: This test is using the Django `TestCase` class instead of being a pytest function like throughout the rest of the codebase. It should be updated to match the rest, as it is now, it's quite difficult to understand what it is checking.

tests/process_manager/views/test_partial_views.py
```diff
@@ -193,3 +196,32 @@ def test_filter_table(search, table, expected):
...
+        def exception_view(request):
+            raise Exception("Test exception")
+
+        request = self.factory.get("/")
+        response = exception_view(request)
+
+        expected_content = render_to_string(
+            "process_manager/partials/error_message.html", request=request
+        )
+        self.assertEqual(response.content.decode(), expected_content)
```
AdrianDAlessandro, line 223: All of this should be achievable with using `assertTemplateUsed` and the `client` fixture.

tests/process_manager/views/test_partial_views.py
```diff
@@ -193,3 +196,32 @@ def test_filter_table(search, table, expected):
...
+
+        mock_logger.exception.assert_called_once()
```
AdrianDAlessandro, line 225: We shouldn't have to mock the logger and be able to use the `caplog` fixture. Although this is probably fine as-is

AdrianDAlessandro (commented):
Sorry I'm a bit late on reviewing this. It appears to do what it is meant to do. It doesn't catch when the message feed is dead (ie when the kafka consumer isn't running), but that doesn't seem to be the reason for it.

I've got a few comments on the test, which I think should be updated

tests/process_manager/views/test_partial_views.py
Sahil590: I can address these comments in my current pr.

Issue #182, opened 2024-10-25
Rerender after each ps makes pagination of processes useless


cvzbynek:
**Bug Summary**:  
When booting 50+ processes, causing the process table to paginate, the UI becomes unresponsive for page navigation. Each `ps` command triggers a re-render, preventing users from switching between pages effectively.

**Steps to Reproduce**:
1. Boot 50+ processes to ensure pagination in the process table.
2. Attempt to navigate to a different page in the paginated table.

**Expected Behavior**:  
User should be able to switch pages smoothly without interruptions, allowing for easy navigation through all process entries.

**Actual Behavior**:  
After each `ps` command, the table re-renders, resetting the pagination and preventing navigation to other pages. This continuous re-rendering disrupts page-switching, making it impossible to view processes beyond the first page.

**Possible Cause**:  
The `ps` command appears to be triggering a full re-render of the process table each time it's executed, possibly due to the way pagination state is handled in relation to re-renders.

**Impact**:  
This behavior hinders usability for users managing large numbers of processes, effectively limiting them to viewing only the initial page of the process list.

![image](https://github.com/user-attachments/assets/658fe3a9-6848-40e6-acd4-41023dc1a6e4)

cc-a:
Thanks @cvzbynek, good catch. I was aware that the current refresh behaviour wouldn't be compatible with pagination but didn't know that django-tables2 would automatically paginate above 50 lines. As we haven't had pagination stated as a clear requirement and we're providing filtering functionality for working with large numbers of entries I'm inclined to simply disable it and see if it comes up in feedback.

cvzbynek:
> Thanks @cvzbynek, good catch. I was aware that the current refresh behaviour wouldn't be compatible with pagination but didn't know that django-tables2 would automatically paginate above 50 lines. As we haven't had pagination stated as a clear requirement and we're providing filtering functionality for working with large numbers of entries I'm inclined to simply disable it and see if it comes up in feedback.

Ok, in next iteration I'll try to disable it.

Issue #189, opened 2024-10-29
 Enhance Table Sorting and Add Advanced Search with Dropdown Filters


cvzbynek:
### Description
Implement enhanced table sorting capabilities and an advanced search bar with a dropdown for filtering by specific parameters. This feature will improve user efficiency by enabling more targeted searches and streamlined sorting within the process manager interface.

### Background
This improvement was suggested by Pierre Lasorak as a desired feature. Adding these elements would align the table functionality more closely with user needs and improve the overall usability of the process management page.

### Proposed Features
1. **Table Sorting**: 
   - Add interactive sorters to table columns, allowing users to sort by fields like UUID, status, and user.
   - Provide options for ascending and descending order to enhance data organization.

2. **Advanced Search with Dropdown Filters**:
   - Implement a search bar with a dropdown menu to filter by specific parameters (e.g., user, status, session).

### Implementation Notes
- **Table Sorters**: Use JavaScript-based sorting for responsiveness and compatibility with the current UI structure.
- **Search Dropdown**: Integrate a dropdown menu that updates the search query dynamically based on selected filters.

@cc-a

cc-a:
Thanks for writing this up @cvzbynek. The search dropdown sounds good, feel free to pick that up if you'd like.

As far as the table sorting goes, minimising javascript on the front-end is a high priority design goal and I don't believe would be compatible with our current approach as subsequent table refreshes would likely clobber any content sorting. Instead I'd suggest using django-tables2 support for sorting. This should be a simple matter of enable sorting for suitable columns and including the current querystring in the `hx-get` attribute of the search input so that sorting is preserved on table refresh. Column sorting will require a page reload but I'd rather minimise complexity and development time unless it's highlighted in user feedback.

cvzbynek:
@cc-a Everything seems to be working as expected in my tests. The JavaScript is kept to a minimum with just a few additional lines. See #207 for details!

PR #207, opened 2024-11-05
Sorters search

cvzbynek:
# Description

This PR introduces sorting capabilities for each column and an advanced search bar, enhancing the table's interactivity and search precision. The updates address the following key improvements:

- **Column Sorting**: Added clickable headers that toggle ascending/descending order for easier data navigation. #189
- **Advanced Search Bar**: Includes a dropdown to specify search parameters, allowing targeted searches by column. #189 
- **Issue Fix**: Resolves issue #182 

## Type of change

- [X] Optimization (non-breaking, back-end change that speeds up the code)
- [X] Bug fix (non-breaking change which fixes an issue)

## Key checklist

- [X] All tests pass (e.g., `python -m pytest`)
- [X] The documentation builds and looks OK (e.g., `python -m sphinx -b html docs docs/build`)
- [X] Pre-commit hooks run successfully (e.g., `pre-commit run --all-files`)

## Further checks

- [X] Code is commented, particularly in hard-to-understand areas
- [ ] Tests added or an issue has been opened to tackle that in the future


dalonsoa (changes requested):
The code looks ok - and certainly works. For sure clearer than having to know that you need to write `column_name:search_term`!

My main concern, and the reason I'm not approving, yet, is because it uses JS to add the interactivity and I think the design choice was to use hyperscript, as we have done in other parts of the tool. I don't mind either way - don't know JS nor hyperscript - but using both is going to be really confusing for developers (in particular future ones).

Having said that, I'm not entirely sure why we need the `combined_search` field at all (and therefore the JS code, to start with). I think it will be way simpler to retrieve the selected column from the dropdown and pass it to the backend and the search function. So, the `hx-include` directive will become: `hx-include="#sort-field, #search-dropdown">`, which I think will pick the selected value from the dropdown. And if it does not, I'm sure there's a way of doing so that do not require scripting in the front end.

As I said, I think this is more a question for discussion than me having specific changes to the code which, otherwise, looks ok.

cvzbynek:
> The code looks ok - and certainly works. For sure clearer than having to know that you need to write `column_name:search_term`!
> 
> My main concern, and the reason I'm not approving, yet, is because it uses JS to add the interactivity and I think the design choice was to use hyperscript, as we have done in other parts of the tool. I don't mind either way - don't know JS nor hyperscript - but using both is going to be really confusing for developers (in particular future ones).
> 
> Having said that, I'm not entirely sure why we need the `combined_search` field at all (and therefore the JS code, to start with). I think it will be way simpler to retrieve the selected column from the dropdown and pass it to the backend and the search function. So, the `hx-include` directive will become: `hx-include="#sort-field, #search-dropdown">`, which I think will pick the selected value from the dropdown. And if it does not, I'm sure there's a way of doing so that do not require scripting in the front end.
> 
> As I said, I think this is more a question for discussion than me having specific changes to the code which, otherwise, looks ok.

Thanks for the feedback! I definitely see your point about keeping things consistent for future developers.

Note that I’m coming from a React background, where a lot of logic is handled in the frontend, so I’m adapting to this approach too.

I’ve made the suggested changes to align better with the code’s overall design, and it does seem more cohesive. Only time will tell if we can manage everything with just hyperscript, but it sounds like a reasonable approach to stick with for now.

process_manager/views/partials.py
```diff
@@ -63,31 +62,39 @@ def process_table(request: HttpRequest) -> HttpResponse:
...
+    if search_dropdown and not search_input:
+        filter_query = ""
+    else:
+        filter_query = (
+            f"{search_dropdown}:{search_input}" if search_input else search_dropdown
         )
-    # Filter table data based on search parameter
-    table_data = filter_table(request.GET.get("search", ""), table_data)
+    # Apply search filtering
+    table_data = filter_table(filter_query, table_data)
```
dalonsoa, line 89: Now that we have the column directly here, we don't need any of this to create a combined search term. Modify the `filter_table` function so it takes the column name as parameter as well and [remove the code within that function](https://github.com/ImperialCollegeLondon/drunc_ui/blob/fcd25490b4834c1c851aac3f9de7078ddbb42713/process_manager/views/partials.py#L38) that precisely split the search query into column and search term.

process_manager/templates/process_manager/index.html
```diff
@@ -171,13 +206,28 @@ <h5>Process Control</h5>
...
+                <input id="search-input"
+                       class="form-control"
+                       type="search"
+                       name="search"
+                       placeholder="Search processes..."
+                       style="flex: 1"
+                       hx-get="{% url 'process_manager:process_table' %}"
+                       hx-trigger="input changed delay:500ms, every 1s"
+                       hx-target="div.table-container"
+                       hx-include="#sort-field, #combined-search, #search-input, #search-dropdown">
```
dalonsoa, line 228: Combined search does not exist anymore, and if it does, it is not needed.
```suggestion
                       hx-include="#sort-field, #search-input, #search-dropdown">
```

process_manager/templates/process_manager/partials/process_table.html
```diff
@@ -1,2 +1,46 @@
...
+                   border-radius: 8px;
+                   font-size: 0.9em;
+                   font-weight: 600;
+                   cursor: pointer;
+                   transition: background-color 0.3s ease"
+            onmouseover="this.style.backgroundColor='#e2e6ea'"
+            onmouseout="this.style.backgroundColor='#f8f9fa'"
+            hx-get="{% url 'process_manager:process_table' %}?sort="
+            hx-target="div.table-container"
+            onclick="document.getElementById('sort-field').value = ''; document.getElementById('sort-param-display').textContent = 'None';">
```
dalonsoa, line 19: I'm not entirely sure - will let @cc-a to comment - but I believe this can be re-written in hyperscript as:
```suggestion
            _="set #sort-field's value to '' and set #sort-param-display's textContent to 'None'">
```

process_manager/templates/process_manager/partials/process_table.html
```diff
@@ -1,2 +1,46 @@
...
+  {% if table.show_header %}
+    <thead class="custom-thead" {{ table.attrs.thead.as_html }}>
+      <tr>
+        {% for column in table.columns %}
+          <th class="custom-th" {{ column.attrs.th.as_html }}>
+            {% if column.orderable %}
+              <a hx-get="{% url 'process_manager:process_table' %}?sort="
+                 hx-target="div.table-container"
+                 href="#"
+                 onclick=" let sortField = document.getElementById('sort-field'); let currentSort = sortField.value; let newSort = currentSort === '{{ column.name }}' ? '-{{ column.name }}' : '{{ column.name }}'; sortField.value = newSort; this.setAttribute('hx-get', `{% url 'process_manager:process_table' %}?sort=${newSort}`); "
```
dalonsoa, line 32: And this one would look:
```suggestion
                 _="let currentSort = #sort-field's value; set #sort-field's value to (if currentSort is '{{ column.name }}' then '-{{ column.name }}' else '{{ column.name }}') and set this's hx-get to '{% url "process_manager:process_table" %}?sort=' + #sort-field's value"
```

I've used Copilot to translate from JS to HS, so don't accept this blindly, but I think it looks sensible.

dalonsoa (changes requested):
This looks much better, but still there are a couple of areas of improvement, both in the front end and in the backend.

cvzbynek:
> This looks much better, but still there are a couple of areas of improvement, both in the front end and in the backend.
Here's your feedback refined and formatted with Markdown:

@dalonsoa @cc-a 

Great points on filtering! 
However, when I tried removing JavaScript, I encountered some odd behaviors, both visually and in the code:

1. **Clear Sorting Button**:  
   After rewriting the clear sorting button in HyperScript, clicking it caused the entire table to reload. This behavior feels excessive for a simple sorting action, where a smoother update would be expected. JavaScript, in this case, seems to handle it more fluidly.

2. **HyperScript and `hx` Parameters**:  
   HyperScript appears to have issues with `hx` parameters, which led to console errors. Sorting by column name didn’t work correctly with HyperScript, and the errors made it challenging to troubleshoot. 

Before I dive into potential workarounds, I wanted to ask—would it be such an issue to retain a bit of JavaScript specifically for sorting?

process_manager/templates/process_manager/partials/process_table.html
cc-a: I don't think that looks like a very valid translation unfortunately.
```suggestion
                 onclick=" let sortField = document.getElementById('sort-field'); let currentSort = sortField.value; let newSort = currentSort === '{{ column.name }}' ? '-{{ column.name }}' : '{{ column.name }}'; sortField.value = newSort; this.setAttribute('hx-get', `{% url 'process_manager:process_table' %}?sort=${newSort}`); "
```

process_manager/templates/process_manager/index.html
```diff
@@ -171,13 +206,28 @@ <h5>Process Control</h5>
...
+                       type="search"
+                       name="search"
+                       placeholder="Search processes..."
+                       style="flex: 1"
+                       hx-get="{% url 'process_manager:process_table' %}"
+                       hx-trigger="input changed delay:500ms, every 1s"
+                       hx-target="div.table-container"
+                       hx-include="#sort-field, #search-input, #search-dropdown">
+              </div>
+              <input type="hidden" id="sort-field" name="sort" value="">
```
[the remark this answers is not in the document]
cvzbynek: Why do you think so? All four lines?

process_manager/templates/process_manager/index.html
```diff
@@ -171,13 +206,28 @@ <h5>Process Control</h5>
...
+                <input id="search-input"
+                       class="form-control"
+                       type="search"
+                       name="search"
+                       placeholder="Search processes..."
+                       style="flex: 1"
+                       hx-get="{% url 'process_manager:process_table' %}"
+                       hx-trigger="input changed delay:500ms, every 1s"
+                       hx-target="div.table-container"
+                       hx-include="#sort-field, #search-input, #search-dropdown">
```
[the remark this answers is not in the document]
cvzbynek: You mean on line 228? And how would I get the values to `GET` ?

cvzbynek:
@cc-a @dalonsoa
The JavaScript has been transcribed to Hyperscript, and there are no errors appearing in the console.

process_manager/templates/process_manager/index.html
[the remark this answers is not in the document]
cc-a: Ah, I was conflating the search and sort functionality. But you should be able to remove #search-input as will be automatically included as coming from the same element.

cc-a (approved):
Latest changes fixed the race condition for me.

dalonsoa (approved):
LGTM

PR #208, opened 2024-11-05
Use web app user in logs and process management

### LINKED ISSUE #111: Use logged in web app user in drunc auth token
Currently drunc is running using a Dummy Authoriser which doesn't actually check the [auth tokens](https://github.com/DUNE-DAQ/druncschema/blob/develop/schema/druncschema/token.proto) used to make gRPC calls. Drunc does however use the user_name value associated with the token.

When we make gRPC calls, we're currently generating tokens using [create_dummy_token_from_unname](https://github.com/DUNE-DAQ/drunc/blob/008c6024cd0c74dfe24e1ecc555752fa6ad5019c/src/drunc/utils/shell_utils.py#L314). This asks the OS for the username of the currently logged in system user (nobody in the case of the docker image). Drunc logs and Kafka messages then show that the system user is performing actions. It would be nicer if the username used in the token corresponded to the logged in user of the web app. It's pretty easy to role our own tokens for this.

TinyMarsh:
# Description
This PR adresses #111 by passing the web app username - from the `request` object available in views - to function signatures where the username is relevant to the function i.e. logs and process management.

Fixes #111

## Type of change

## Key checklist

## Further checks


cc-a (changes requested):
Looks good. One tiny tweak so that our dummy processes can keep running.

dalonsoa (approved):
Keeping Chris comment aside, this looks good to me!

process_manager/process_manager_interface.py
```diff
@@ -6,32 +6,33 @@
...
     ProcessUUID,
 )
+from druncschema.token_pb2 import Token
 
 
-def get_process_manager_driver() -> ProcessManagerDriver:
+def get_process_manager_driver(username: str) -> ProcessManagerDriver:
     """Get a ProcessManagerDriver instance."""
-    token = create_dummy_token_from_uname()
+    token = Token(token=f"{username}-token", user_name=username)
```
AdrianDAlessandro, line 21: Assuming this actually works with their authentication system

AdrianDAlessandro (approved):
LGTM!

process_manager/process_manager_interface.py
cc-a: For the moment they're using a `DummyAuthoriser` that doesn't check the token.

cc-a (approved):

PR #213, opened 2024-11-06
State machine

### LINKED ISSUE #161: Finite State Machine Representation - PoC
Requires #160.

Whilst we eventually want a nicely laid out and rendered representation of the Finite State Machine (FSM) we're going to focus here on the fundamental functionality and prettify things later. For this current work we can assume the FSM will be as documented in the drunc wiki - https://github.com/DUNE-DAQ/drunc/wiki/FSM - however future work may consider working with arbitrary FSM's that will be loaded from configuration.

This issue proposes a simple representation of the FSM:
- a row of labels corresponding to the states of the FSM. The current state of the root controller (as reported by the gRPC interface) is highlighted in a different colour.
- a row of buttons corresponding to the transitions of the FSM. The currently available transitions are enabled, all others are disabled. Clicking a button triggers a view function that executes the transition on the root controller (via the gRPC interface) then reloads the page.

You'll need to add relevant view functions and wrapper functions to the controller driver interface module (see #160) as you go.

## Breakdown (by @dalonsoa )

This issue is actually bigger than planned, so breaking it down into smaller chunks:

- [x] #215
- [x] #216
- [x] #217

### LINKED ISSUE #215: Implement simple, clickable representation of the FSM
Implements the front end of the FSM, with buttons/labels indicating the states and clickable transitions that the user can use to move from one state to another. The layout should be whatever is useful, but in principle it should represent the FSM described in the https://github.com/DUNE-DAQ/drunc/wiki/FSM .

The FSM structure cannot be retrieved in full from the Controller, so the states and the transitions will need to be hardcoded - althgouh the front end representation should be flexible enough so that using a different FSM will still render fine.  

No actual link with the backend is to be done at this time.

dalonsoa:
# Description

This PR implements the proof of concept of the state machine view. 

Currently, the state machine can only go from `initial` to `configured` and back. Running `start` requires to pass arguments to the command, which is implemented in #227 

Fixes #215 
Partly addresses #217 

https://github.com/user-attachments/assets/1f41973f-5c38-4ae7-8725-0b26bcb488e3

## Type of change

## Key checklist

## Further checks


controller/fsm.py
```diff
@@ -0,0 +1,63 @@
...
+        """
+        return [
+            "None",
+            "Initial",
+            "Configured",
+            "Ready",
+            "Running",
+            "Dataflow_Drained",
+            "Triggered_Sources_Stopped",
+        ]
```
cc-a, line 63: This is a good stand in for the controller interface but we shouldn't need it once we have that interface working otherwise we risk duplicating the state of the controller.

cc-a (commented):
Looks good, a nice start to things.

controller/fsm.py
dalonsoa: Well, yes and no. Based on the conversations with Pierre, out of the controller you can get:
- The current state
- The transitions allowed from this state
- All the possible transitions

But you cannot get:
- All possible states
- To what state each transition leads you

So it is not possible to draw a map of the FSM out of the information facilitated by the Controller - we need to hardcode it somehow.

Having said that, I totally agree that the Controller should be the single source of truth of the current state, so some of the features of this structure are either redundant or will not be used. You can be seen more as an aid to the representation of the machine, if you want, but the state is not really hold by it, but by the Controller.

controller/controller_interface.py
```diff
@@ -12,6 +12,33 @@ def get_controller_driver() -> ControllerDriver:
...
+
+    Args:
+        event (str): The event to send.
+        **kwargs (dict[str, str]): The arguments for the event.
+
+    Returns:
+        FSMResponseFlag: The flag returned by the controller. 0 if the event was
+            successful, 1-4 if the event failed.
+    """
+    return get_controller_driver().execute_fsm_command(event, **kwargs).data.flag
```
cc-a, line 44: I didn't think this PR did anything with the controller backend?

controller/fsm.py
```diff
@@ -0,0 +1,109 @@
...
+
+    # The transitions between the states
+    boot = none.to(initial)
+    terminate = initial.to(none)
+    conf = initial.to(configured)
+    scrap = configured.to(initial)
+    start = configured.to(ready)
+    enable_triggers = ready.to(running)
+    disable_triggers = running.to(ready)
+    drained_dataflow = ready.to(dataflow_drained)
```
cc-a, line 28: Should this maybe be:
```suggestion
    drain_dataflow = ready.to(dataflow_drained)
```

controller/fsm.py
```diff
@@ -0,0 +1,109 @@
+"""Module that implements `drunc` finite state machine."""
+
+from statemachine import State, StateMachine
+
+from . import controller_interface as ci
+
+
+class DruncFSM(StateMachine):
```
cc-a, line 8: I'm still unsure what we'll need this class for once we have the controller backend. It's a cute way to define things but my impression is we only need to store a mapping of state names to transitions and transitions to targets.

controller/fsm.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+        return [
+            "None",
+            "Initial",
+            "Configured",
+            "Ready",
+            "Running",
+            "Dataflow_Drained",
+            "Triggered_Sources_Stopped",
+        ]
```
cc-a, line 82: Does the class not store these under under a collection somewhere? That might just follow the order they are defined in.

controller/templates/controller/partials/state_machine.html
```diff
@@ -0,0 +1,10 @@
+{% load render_table from django_tables2 %}
+{% for event in events %}<input type="hidden" name="event" id="event_{{ event }}" value="{{ event }}">{% endfor %}
```
cc-a, line 2: I'm found it a little difficult to puzzle out how this worked. Could we not use the same approach as the action buttons in the process manager. Keeps the behaviour attached to the buttons that are clicked.

controller/templates/controller/partials/state_machine.html
```diff
@@ -0,0 +1,10 @@
...
+<form>
+  {% csrf_token %}
+  <input type="hidden"
+         name="current_state"
+         id="current_state"
+         value="{{ current_state }}">
```
cc-a, line 8: Maybe also add a comment here that this can be removed later.

cc-a (changes requested):
Thanks Diego. Looks very nice in a table like that. I've added a few questions. One thing that needs fixing is that the lock file is out of sync with pyproject.toml so the docker image doesn't build. This wasn't caught by CI because we're using an old poetry version there that only gives a warning. We could do to update our CI to stop using the poetry setup action anyway so I'll cerate an issue for that.

controller/controller_interface.py
```diff
@@ -12,6 +12,33 @@ def get_controller_driver() -> ControllerDriver:
...
+    """
+    return get_controller_status().state
+
+
+def send_event(event: str, **kwargs: dict[str, str]) -> FSMResponseFlag:
+    """Send an event to the controller.
+
+    Args:
+        event (str): The event to send.
+        **kwargs (dict[str, str]): The arguments for the event.
```
cc-a, line 38: No need to repeat the type info in the docstring. Mkdocs can get it from the type hint.

controller/fsm.py
dalonsoa: I've simplified things massively using simple mappings. You were definitely right there.

controller/templates/controller/partials/state_machine.html
dalonsoa: Removed

controller/templates/controller/partials/state_machine.html
dalonsoa: Simplified

dalonsoa:
This PR has been massively simplified using mappings, as you suggested, @cc-a . Definitely the right call.

cc-a (approved):
Nice. Seems to have integrated with the backend fairly smoothly.

I couldn't progress past the `configured` state. From the error message I assume we need #216 to progress any further through the fsm.

dalonsoa:
> I couldn't progress past the `configured` state. From the error message I assume we need #216 to progress any further through the fsm.

Yep, that's explained in the PR description ;)

tests/controller/views/test_partial_views.py
```diff
@@ -0,0 +1,48 @@
...
+from django.urls import reverse
+
+from controller import fsm
+from controller.tables import FSMTable
+
+from ...utils import LoginRequiredTest
+
+
+class TestFSMView(LoginRequiredTest):
+    """Test the process_manager.views.process_table view function."""
```
dalonsoa, line 14:
```suggestion
    """Test the controller.views.state_machine view function."""
```

jamesturner246 (approved):
It looks really nice and fancy. LGTM! :+1: 

There should probably be a timer based refresh for the FSM pane though, like in the other partials.

controller/templates/controller/index.html
```diff
@@ -13,8 +13,36 @@
...
+    <div class="flex-container" id="main-content">
+      <!-- Finite state machine Panel -->
+      <div class="fst-panel">
+        <div class="card shadow-sm rounded">
+          <div class="card-header bg-primary text-white rounded-top mb-3">
+            <h5>Finite State Machine</h5>
+          </div>
+          <div class="card-body">
+            <div hx-get="{% url 'controller:state_machine' %}"
+                 hx-trigger="load"
```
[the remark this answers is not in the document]
dalonsoa: Maybe. That will not play well with the dialog (in #241) if it is open when refreshing, though. For now, I'm tempted to leave this to the old "refresh whole page approach with CTRL+R", and maybe open an issue about it.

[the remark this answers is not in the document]
dalonsoa: I've just opened an issue about this. https://github.com/ImperialCollegeLondon/drunc_ui/issues/257

AdrianDAlessandro (approved):
I haven't analysed the code in detail, but it seems to work reasonably well.

I see a weird bit of behaviour with if I try to kill the controller processes in the the process manager I can't then re-boot them without taking the whole stack down and up again, but I don't think that's related to this PR.

dalonsoa:
It's not to do with this PR, but it is true that error handling is not great. If something bad happens to the controller, the FSM starts doing strange things like not reporting any change in state despite not raising an error, either, and things like that. But I don't know if that is an artefact of our "fake" drunc setup or if it will work like that in practice. Might be worth creating an issue to explore the options in the future.

PR #218, opened 2024-11-08
Add severity data to broadcast message feed

### LINKED ISSUE #204: Add severity data to broadcast message feed
There is some structured data about the purpose/severity of messages sent via drunc. This is contained [BroadcastMessage.type] which contains  an enum choice of [BroadcastType]. This severity informtaion should be presented to the user in the message feed. Some input from the DUNE team on how they would like the various enum values to be rendered. For instance, it might make sense to map the possible enum values to DEBUG, ERROR, INFO levels.

Implementing this could explore changing the message feed to a table rather than an unordered list.

Sahil590:
# Description

This adds severity information from the messages received.
Fixes #204  (issue)

## Type of change

## Key checklist

## Further checks


dalonsoa (approved):
The severity-related code looks reasonable to me 👍 !

main/management/commands/kafka_consumer.py
```diff
@@ -6,11 +6,32 @@
...
+    BroadcastType.DRUNC_EXCEPTION_RAISED: "ERROR",
+    BroadcastType.UNHANDLED_EXCEPTION_RAISED: "CRITICAL",
+    BroadcastType.STATUS_UPDATE: "INFO",
+    BroadcastType.SUBPROCESS_STATUS_UPDATE: "INFO",
+    BroadcastType.DEBUG: "DEBUG",
+    BroadcastType.CHILD_COMMAND_EXECUTION_START: "INFO",
+    BroadcastType.CHILD_COMMAND_EXECUTION_SUCCESS: "INFO",
+    BroadcastType.CHILD_COMMAND_EXECUTION_FAILED: "ERROR",
+    BroadcastType.FSM_STATUS_UPDATE: "INFO",
+}
```
cc-a, line 33: One thing I noticed with this current mapping is that we get a mix of INFO and DEBUG statuses for `ps` messages. So "attempting to execute ps" is debug but "successfully executed ps" is info. Seems more consistent if they're both debug. Could you find out  BroadcastType value is being received on the success messages?

main/models.py
```diff
@@ -24,3 +24,4 @@ class DruncMessage(models.Model):
     topic = models.CharField(max_length=255)
     timestamp = models.DateTimeField()
     message = models.TextField()
+    severity = models.CharField(max_length=10, default="INFO")
```
cc-a, line 27: Would be nice to use [choices](https://docs.djangoproject.com/en/5.1/ref/models/fields/#choices) here to limit the possible values of the field.

main/tables.py
```diff
@@ -0,0 +1,11 @@
+"""Defines the Drunc Message Table for the data from the Kafka messages."""
+
+import django_tables2 as tables
+
+
+class DruncMessageTable(tables.Table):
+    """Defines a Drunc Message Table for the data from the Kafka messages."""
+
+    timestamp = tables.DateTimeColumn(format="y-m-d , H:i")
```
cc-a, line 9: The sorting of this isn't compatible with the current htmx setup so we should disable it unless it's requested as a feature.

main/views/partials.py
```diff
@@ -4,29 +4,27 @@
...
-    )
+    ).order_by("-timestamp")
 
-    # Time is stored as UTC. localtime(t) converts this to our configured timezone.
-    messages = [
-        f"{localtime(record.timestamp).strftime('%Y-%m-%d %H:%M:%S')}: {record.message}"
-        for record in records
-    ]
+    table = DruncMessageTable(records)
+    RequestConfig(request).configure(table)
```
cc-a, line 24: This will by default paginate the table which is also not compatible with how we're using htmx here. I think you can do 
```suggestion
    table = DruncMessageTable(records)
    RequestConfig(request, paginate=False).configure(table)
```
to disable.

cc-a (changes requested):
Thanks @Sahil590. Looks good, just a couple of tweaks.

main/management/commands/kafka_consumer.py
Sahil590: The two BroadcastType values that were returned are 0 and 6. I think I may have a fix with, but having some issues with my setup, so can't test anything.

Sahil590: I addressed the comments below since they are straight forward

main/management/commands/kafka_consumer.py
```diff
@@ -6,11 +6,32 @@
...
 from django.conf import settings
 from django.core.management.base import BaseCommand
-from druncschema.broadcast_pb2 import BroadcastMessage
+from druncschema.broadcast_pb2 import BroadcastMessage, BroadcastType
 from kafka import KafkaConsumer
 
 from ...models import DruncMessage
 
+BROADCAST_TYPE_SEVERITY = {
+    BroadcastType.ACK: "INFO",
```
cc-a, line 15: One of the stated use cases was that they'd like to be able to filter out the constant ps messages that are sent so I think this does need to be debug and let's try out making `COMMAND_EXECUTION_START` debug as well.

cc-a (approved):
Thanks Sahil. Looks nice. Neat in the UI as well. Could you just update the severity mapping as I've indicated.

AdrianDAlessandro (approved):
This looks good! My only critique is that the messages now render with a horizontal scroll bar because the Messages take up a lot of horizontal space and do not wrap within the table. Not sure where in the templates this should be changed, but it would be more user-friendly if it wrapped.

Sahil590:
> This looks good! My only critique is that the messages now render with a horizontal scroll bar because the Messages take up a lot of horizontal space and do not wrap within the table. Not sure where in the templates this should be changed, but it would be more user-friendly if it wrapped.

I don't get the horizontal scrolling, does it cut off a lot of the message on your screen?

AdrianDAlessandro:
> > This looks good! My only critique is that the messages now render with a horizontal scroll bar because the Messages take up a lot of horizontal space and do not wrap within the table. Not sure where in the templates this should be changed, but it would be more user-friendly if it wrapped.
> 
> @AdrianDAlessandro I don't get the horizontal scrolling, does it cut off a lot of the message on your screen?

Yes, any of the message that is longer than the standard `ps` message gets cut off

Screenshot 2024-11-20 at 12 56 00
Screenshot 2024-11-20 at 12 56 12

Issue #140, opened 2024-10-08
Look into asynchronous views


cc-a:
We are currently using standard syncronous views in Django. This leads to us needing to wrap the asyncronous interface to the DRUNC process manager.

It would be nice to use asynchronous views in order to:
- Reduce the boiler plate of wrapping methods of the `ProcessManagerDriver` class.
- Exploit performance benefits. This especially may be useful as we expect the multiple simultaneous frontends to be polling our server and asyncronous views would help defray the performance implications of interacting with the process manager.

Need to have a careful look at this as we've not done it before and there are several caveats noted in the Django docs - https://docs.djangoproject.com/en/5.1/topics/async/.

cc-a:
Just stumbled across [Daphne](https://docs.djangoproject.com/en/5.1/howto/deployment/asgi/daphne/) which looks like it will be useful for testing async during development.

cc-a:
Just noting that we have at least piece of sync only middleware - whitenoise. There is a whitenoise fork called [ServeStatic](https://archmonger.github.io/ServeStatic/latest/) which has async support and looks promising.

cc-a:
This blog is an interesting read on the topic https://jonathanadly.com/is-async-django-ready-for-prime-time

TinyMarsh:
After exploring this issue and some discussion, we decided that while async views could reduce boilerplate and improve performance, we've decided to put this work on hold for now.

The primary blocker is the non-async nature of Django's template rendering system. Every call to render a template would need to be wrapped in sync_to_async, introducing complexity and potentially negating some of the performance benefits we hoped to gain.

Until Django templates natively support async or we find a way to address this without adding significant overhead, we're shelving this effort. We'll revisit this at a later date, where we could explore alternatives such as third-party async template engines like Jinja2 (which Django supports natively).

PR #219, opened 2024-11-08
Asynchronous `process_table` view

TinyMarsh:
# Description

Whilst addressing #140, I came across some problems that were slightly more tricky to solve than others. So in the mean time whilst I'm working through those, I thought I would create this small PR which just addresses the `process_table` view just to make sure what I'm doing is sensible.

* `process_manager/process_manager_interface.py`: Removed the `_get_session_info` function and refactored `get_session_info` to directly use asynchronous calls without `asyncio.run`.
* `process_manager/views/partials.py`: Converted the `process_table` view to an asynchronous function.
* `tests/process_manager/views/test_partial_views.py`: Updated the `test_get` and `test_get_with_search` methods in `TestProcessTableView` to be asynchronous.

## Type of change

- [x] Optimization (non-breaking, back-end change that speeds up the code)

## Key checklist

## Further checks

### DIFF
--- a/process_manager/views/partials.py
+++ b/process_manager/views/partials.py
@@ -1,6 +1,7 @@
 """View functions for partials."""
 
 import django_tables2
+from asgiref.sync import async_to_sync, sync_to_async
 from django.contrib.auth.decorators import login_required
 from django.http import HttpRequest, HttpResponse
 from django.shortcuts import render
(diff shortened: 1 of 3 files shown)


process_manager/views/partials.py
```diff
@@ -50,17 +51,18 @@ def filter_table(
     return [row for row in table if any(search in str(row[k]).lower() for k in columns)]
 
 
+@sync_to_async
 @login_required
 @handle_errors
-def process_table(request: HttpRequest) -> HttpResponse:
+@async_to_sync
+async def process_table(request: HttpRequest) -> HttpResponse:
```
dalonsoa, line 58: Assuming I got this right from the [Django docs on the topic](https://docs.djangoproject.com/en/5.1/topics/async/), we need this stack of decorators because `@login_required` and `@handle_errors` need to run in the main event loop, right? But except that bit - before entering into the view or if getting out of it unexpectedly because of an error - the view itself runs asyncronously. Is this correct?

dalonsoa (approved):
The code looks ok, as far as I can understand it, but it is unclear to me if going in and out of the async path has any performance penalty that will defeat the purpose of running things asynchronously.

PR #225, opened 2024-11-11
Enable `drunc` compose profile to build on MacOS

AdrianDAlessandro:
# Description

This PR makes the full `drunc` compose profile work on MacOS. Previously there was an error about a `SPACK_ROOT` environment variable not being set. Thanks to @cc-a for helping find what that environment variable should be.

**Note:** This makes changes to the Dockerfile, so it should be tested on all different operating systems before merging. Tick the box if it works for you:
- [x] MacOS
- [ ] Windows
- [x] Linux

Fixes #209 

## Type of change

- [ ] Documentation (non-breaking change that adds or improves the documentation)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (non-breaking, back-end change that speeds up the code)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (whatever its nature)

## Key checklist

- [ ] All tests pass (eg. `python -m pytest`)
- [ ] The documentation builds and looks OK (eg. `python -m sphinx -b html docs docs/build`)
- [ ] Pre-commit hooks run successfully (eg. `pre-commit run --all-files`)

## Further checks

- [ ] Code is commented, particularly in hard-to-understand areas
- [ ] Tests added or an issue has been opened to tackle that in the future. (Indicate issue here: # (issue))

PR #233, opened 2024-11-14
Adopt nightly image for controller support

### LINKED ISSUE #203: Investigate use of a nightly build for controller image
The current base image used for the full drunc stack comes from a release some time ago. There are some key limitations due to its age that make it unsuitable for developing the controller image. In order to make progress with the controller UI development we will likely have to start using a nightly build (primarily to take advantage of the pending recursive status functionality). This was briefly explored when developing the controller image but dropped in favour of using a release build. The process to adopt a nightly image should be something like:

- Update the base image for the full drunc stack to a dune-daq nightly.
- Troubleshoot the docker stack until we can spawn the controller test session (here be dragons, will required support from DUNE team).
- Upload a copy of the nightly image to ghcr under our own repository.
- Update the base image to use our copy.

cc-a:
# Description

This PR adopts a nightly build of the dune dev environment. This been pinned and uploaded to a package linked to this repository which we will use as the basis for development going forward.

To try this out you should be able to:
```bash
docker compose --profile drunc up --build
docker compose exec drunc /boot_test_session.sh
docker compose exec app python scripts/talk_to_controller.py
```

You should see a tree output showing all controller/process statuses.

Some reasonably significant changes have happened on the drunc side. Most notably the controllers no longer have fixed ports but instead you need to query the connectivity service to find the uri where components are running. See `scripts/talk_to_controller.py` for how to do this. Once merged we'll need to implement this behaviour for the web app.

Fixes #203  (issue)

## Type of change

## Key checklist

## Further checks


dalonsoa (approved):
This works exactly as described 👍 So, from my point of view, this is ready to go.

pyproject.toml
```diff
@@ -11,14 +11,15 @@ authors = [
 python = "^3.11"
 django = "^5.1"
 whitenoise = "^6.8.2"
-drunc = { git = "https://github.com/DUNE-DAQ/drunc.git" }
-druncschema = { git = "https://github.com/DUNE-DAQ/druncschema.git" }
+drunc = { git = "https://github.com/DUNE-DAQ/drunc.git", rev = "f068eaccb840da03a104adba2da3f27454c218d5" }
+druncschema = { git = "https://github.com/DUNE-DAQ/druncschema.git", rev = "a07226fa9c3a4329b4c38e084c913cb1e2db0b4e" }
```
AdrianDAlessandro, line 15: This means it relies on a fixed version of drunc and druncschema and won't pick up on any future updates. Is this intentional so that it matches the nightly build we have and updated versions will only be brought in after they're tested alongside this?

drunc_docker_service/Dockerfile
```diff
@@ -12,13 +12,12 @@ WORKDIR /basedir
 # See https://github.com/DUNE-DAQ/drunc/wiki/Setup-drunc-with-DUNE-DAQ for explanation
 RUN source /cvmfs/dunedaq.opensciencegrid.org/setup_dunedaq.sh && \
         setup_dbt latest_v5 && \
-        dbt-create fddaq-v5.1.0-a9 fddaq-v5.1.0-a9 && \
-        cd fddaq-v5.1.0-a9 && \
+        dbt-create -n NFD_DEV_241114_A9 && \
+        cd NFD_DEV_241114_A9 && \
```
AdrianDAlessandro, line 16: Seems like this PR was made before the change in #225 
We will need to add the equivalent line from that PR now that this path has changed

AdrianDAlessandro (commented):
This wasn't up to date with main so it has broken the fix in #225

pyproject.toml
cc-a: Yeah exactly. We want the version the app is using to exactly match the development environment. It's nightly build so that's just some random commit. Hopefully we'll be able to move to a new release at some point.

drunc_docker_service/Dockerfile
dalonsoa: Yep. That branch was up to date with `main`, incorporated the change in #225

PR #238, opened 2024-11-19
Message filter

### LINKED ISSUE #205: Advanced filtering of message feed
Needs #204.

The addition of severity data to the message feed will allow support for more advanced filtering of messages. We should support:

- filtering by severity value (depending on implementation of #204) possibly by a drop down of available values.
- free text filtering of message content.

Sahil590:
# Description
Adding advanced message feed filtering using htmx the same way we currently use it for the search. 
Fixes #205  (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/main/templates/main/messages.html
+++ b/main/templates/main/messages.html
@@ -13,13 +13,31 @@
     <div class="card-body">
       {% now "e (O)" as server_time %}
       <span id="timezone" class="d-line text-body-tertiary">All timestamps are displayed in {{ server_time }}</span>
-      <input class="form-control mb-2"
-             type="search"
-             name="search"
-             placeholder="Search messages..."
-             hx-get="{% url 'main:messages' topic|default:"ALL" %}"
-             hx-trigger="input changed delay:500ms, every 1s"
-             hx-target="#message-list">
+      <!-- Filter Form -->
+      <form id="filter-form" method="get">
+        <!-- Search Input -->
+        <input class="form-control mb-2"
+               type="search"
+               name="search"
+               placeholder="Search messages..."
+               hx-get="{% url 'main:messages' topic|default:"ALL" %}"
+               hx-trigger="input changed delay:500ms, every 1s"
+               hx-target="#message-list"
+               hx-include="#filter-form">
+
+        <!-- Severity Filter Dropdown -->
+        <select name="severity" class="form-select mb-2"
+                hx-get="{% url 'main:messages' topic|default:"ALL" %}"
+                hx-trigger="change delay:500ms"
+                hx-target="#message-list"
+                hx-include="#filter-form">
+          <option value="">All Severities</option>
+          <option value="DEBUG">DEBUG</option>
+          <option value="INFO">INFO</option>
+          <option value="ERROR">ERROR</option>
+          <option value="CRITICAL">CRITICAL</option>
+        </select>
+      </form>
       <div id="message-list"
            class="list-group"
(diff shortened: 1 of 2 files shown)


Sahil590:
@cc-a I have addressed the severity filtering and the free text filtering is done by search, is there anything else that should be added?

main/templates/main/messages.html
```diff
@@ -13,13 +13,31 @@ <h5 class="d-inline">Messages</h5>
...
+               placeholder="Search messages..."
+               hx-get="{% url 'main:messages' topic|default:"ALL" %}"
+               hx-trigger="input changed delay:500ms, every 1s"
+               hx-target="#message-list"
+               hx-include="#filter-form">
+        <!-- Severity Filter Dropdown -->
+        <select name="severity"
+                class="form-select mb-2"
+                hx-get="{% url 'main:messages' topic|default:"ALL" %}"
+                hx-trigger="change delay:500ms"
```
cc-a, line 31: don't think we need the delay here

cc-a (approved):
Looks good @Sahil590. One tiny suggested tweak to make it more responsive. We may need to consider a more sophisticated way to select multiple severities at once but lets get some user feedback first.

AdrianDAlessandro (approved):
I agree with Chris, being able to filter multiple severities (or filter _out_ individual ones) will probably be necessary, but this puts in an initial pass that we can get some feedback on.

main/templates/main/messages.html
Sahil590: Yeah, I did try to play around with adding multiple but would need to add some custom JavaScript for that

Sahil590: I tried out Django filter but was having issues with the constant refreshing of the message pane. I could have found a workaround with custom JavaScript, but I thought this was cleaner.

main/templates/main/messages.html
```diff
@@ -13,13 +13,30 @@ <h5 class="d-inline">Messages</h5>
...
+        <input class="form-control mb-2"
+               type="search"
+               name="search"
+               placeholder="Search messages..."
+               hx-get="{% url 'main:messages' topic|default:"ALL" %}"
+               hx-trigger="input changed delay:500ms, every 1s"
+               hx-target="#message-list"
+               hx-include="#filter-form">
+        <!-- Severity Filter Dropdown -->
+        <select name="severity"
```
dalonsoa, line 28: Far from being an expert in HTML, you should be able to select multiple options here by adding the `multiple` attribute. See https://www.w3schools.com/TAGS/att_select_multiple.asp Now idea what the choice look like in the backend, but you should be able to handle al the required logic there.

dalonsoa (commented):
I think it should be pretty easy to handle multiple severities - see my comment. _Excluding_ certain severities is another story.

main/templates/main/messages.html
dalonsoa: Ok, it won't work exactly as a dropdown, but can be tweaked a bit. See https://stackoverflow.com/a/65969350/3778792

cc-a: If we go this route then [Shoelace dropdowns](https://shoelace.style/components/dropdown) support having checkboxes. We'd need to figure out how it interacts with HTMX though.

jamesturner246:
LGTM. For multiple severities, regex filtering could work: `^(debug|critical)$` etc, like in kafka topic: 

```python
regex = f"^({'|'.join(severities_to_show)})$"
DruncMessage.objects.filter(severity__regex=regex, ...
```

PR #244, opened 2024-11-25
Asynchronous views

### LINKED ISSUE #140: Look into asynchronous views

TinyMarsh:
# Description

_Please include a summary of the change and which issue is fixed (if any). Please also
include relevant motivation and context. List any dependencies that are required for
this change._

Fixes #140 

## Type of change

## Key checklist

## Further checks


TinyMarsh:
Closing this given we have closed the issue relating to this work: #140

PR #248, opened 2024-11-29
Fixes bug in test_get_with_search

### LINKED ISSUE #155: [BUG] Test for table partial views test nothing
Most tests in the [TestProcessTableView](https://github.com/ImperialCollegeLondon/drunc_ui/blob/b10e54e6c2a44852616dc97947e542f60c3eb0d7/tests/process_manager/views/test_partial_views.py#L14) have checks like:

```python
        for row in table.data.data:
            assert row["checked"] == (row["uuid"] in selected_uuids)
```

The problem is that `table.data.data` actually is an empty list, so none of these `assert` statements are run. To check that, just add a `breakpoint()` in front of `assert`: it is never called. 

I've no idea why this is the case, but I suspect that the `ProcessTable` [here](https://github.com/ImperialCollegeLondon/drunc_ui/blob/b10e54e6c2a44852616dc97947e542f60c3eb0d7/process_manager/views/partials.py#L45) cannot make a table out of a list of dictionary full of mocks. As far as I can tell, there's no warning.

Or, more likely, [this mock here](https://github.com/ImperialCollegeLondon/drunc_ui/blob/b10e54e6c2a44852616dc97947e542f60c3eb0d7/tests/process_manager/views/test_partial_views.py#L34) is not working as it should because if we check the contents of `table_data` before calling `ProcessTable`, such table is empty, suggesting that the ProcessSessions is not... iterating.

Whatever the reason, something is not right in these tests.

TinyMarsh:
# Description

This PR fixes a bug in the `test_get_with_search` test, where `table_data` in the `process_table` view was not being update correctly due to incorrect mocking of the session data. The mock for `get_session_info` now correctly simulates the "session" info, and the test has been updated to properly verify the filtered data.

Fixes #155

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/process_manager/views/test_partial_views.py
+++ b/tests/process_manager/views/test_partial_views.py
@@ -33,7 +33,7 @@ class TestProcessTableView(LoginRequiredTest):
         sessions = sessions or [f"session{i}" for i in range(len(uuids))]
         for instance_mock, uuid, session in zip(instance_mocks, uuids, sessions):
             instance_mock.uuid.uuid = str(uuid)
-            instance_mock.session = session
+            instance_mock.process_description.metadata.session = session
             instance_mock.status_code = 0
         mock().data.values.__iter__.return_value = instance_mocks
         return mock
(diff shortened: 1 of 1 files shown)


dalonsoa (approved):
LGTM, and checked it works as expected.

cc-a (approved):

PR #259, opened 2024-12-04
:bug: Manually add SIGHUP attr during tests.

dalonsoa:
# Description

Conditionally adds `SIGHUP` attribute to the `signal` module when running tests in Windows. Apparently, it was removed in Python 3.7 from the Windows version. I don't know if this hack is also needed in MacOS. If it is, it should be easy to fix.

More info in https://stackoverflow.com/a/65906782/3778792

Fixes # N/A

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,1 +1,7 @@
 """Unit tests for drunc_ui."""
+import platform
+import signal
+
+# Windows does not have SIGHUP, so we define it here to avoid errors during tests
+if platform.system() == 'Windows':
+    signal.SIGHUP = 1  # type: ignore[attr-defined]


dalonsoa:
`mypy` is annoying...

dalonsoa:
Doesn't work as a fixture, even with autouse. I think it is because things that require SIGHUP are imported outside of the tests, like in [here](https://github.com/ImperialCollegeLondon/drunc_ui/blob/271bdc677fa22c4207de0997cef91d084fbdda3c/tests/test_process_manager_interface.py#L5), so the fixtures have no effect. If all of those imports happen within a test, then yes, it should be ok.

dalonsoa:
Happy to move all the imports that currently are in the wrong place to within the tests, with the corresponding duplication of import statements, if you think that's the best way forward.

cc-a:
Just to check is it any better if you use the `session` scope for the fixture?

dalonsoa:
Nope. Same problem.

cc-a:
Or possibly the [pytest_configure](https://docs.pytest.org/en/stable/reference/reference.html#pytest.hookspec.pytest_configure) hook.

dalonsoa:
Not the hook, either.

cc-a:
Damn... Does the current implementation work if you do e.g. `pytest main`? I think if it's not trying to run anything in the tests directory it won't import `tests/__init__.py` so you won't get the fix. Not the biggest issue in the world but something to be aware of/document when working in Windows.

dalonsoa:
No, there I have another error, related to Kafka: `E   ModuleNotFoundError: No module named 'kafka.vendor.six.moves'` . Essentially, I can only run `poetry run pytest tests` (assuming I have this fix in place)

cc-a:
Are you sure you're using python 3.11? I think that error is the reason we can't use python 3.12.

dalonsoa:
🤦

dalonsoa:
Since in `pyproject.toml` it says:

```
[tool.poetry.dependencies]
python = "^3.11"
```

My system just picked the default python version, which is 3.12. If this does not work with python 3.12, I really think we should pin the python version in pyproject.toml to 3.11.

dalonsoa:
After using the right python version, `pytest main` works fine.

cc-a (approved):
I can't seem to come up with anything better and it seems to work so lets go ahead with it.

cc-a:
The `python = "^3.11"` should pin to any 3.11.X.

dalonsoa:
I don't know what it is for, but it is used by `drunc` here: https://github.com/DUNE-DAQ/drunc/blob/1848d8494d5583e1a551134d3cd3a38aea810cd0/src/drunc/utils/utils.py#L254

So anything that imports `drunc.utils.utils`, which our code does indirectly by importing several drunc modules, will be hit but the error if `SIGHUP` does not exist.

dalonsoa:
It is funny, because a few lines below it checks if the platform is `linux` as that function will only work in that platform, but even just declaring the function with that default value will fail if not in linux (well, and in Mac, it would seem).

cc-a:
> Seems like it's not needed on MacOS. But a quick question - what is `SIGHUP` and why is it needed?

Long story short it's a *nix OS feature used by drunc relating to management of processes.

Slightly longer version is that SIGHUP is the hangup signal. One of a set of signals that can be sent to running processes by an OS as defined by the POSIX standard. These do all sorts of things that you want to be able to do to processes e.g. the way that Ctrl+C kills processes in the terminal is by sending a signal (SIGINT) to the running subprocess. More reading on wikipedia of course - https://en.wikipedia.org/wiki/Signal_(IPC)#POSIX_signals.

PR #260, opened 2024-12-04
Consolidate use of nightly tag in docker image

cc-a:
# Description

This PR removes some of the duplication of the version tag used by the nightly images. This will simplify updates as it will only have to be changed in one place.

Fixes # (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/drunc_docker_service/Dockerfile
+++ b/drunc_docker_service/Dockerfile
@@ -6,19 +6,20 @@ RUN yum install -y openssh-server && yum clean all
 SHELL ["/bin/bash", "-c"]
 
 WORKDIR /basedir
+ENV NIGHTLY_TAG=NFD_DEV_241114_A9
 
 # setup for the the dune development environment that provides the full set of
 # dependencies required for booting of the controller test session
 # See https://github.com/DUNE-DAQ/drunc/wiki/Setup-drunc-with-DUNE-DAQ for explanation
-ENV SPACK_ROOT /cvmfs/dunedaq-development.opensciencegrid.org/nightly/NFD_DEV_241114_A9/spack-0.22.0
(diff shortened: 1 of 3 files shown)


dalonsoa (approved):
Looks good to me. Simpler, indeed.

The fact that we are using as base image something labelled `nigthly-241114` has nothing to do with the actual nightly build we are using?

drunc_docker_service/Dockerfile
```diff
@@ -6,19 +6,20 @@ RUN yum install -y openssh-server && yum clean all
...
-        dbt-build && \
-        dbt-workarea-env
-WORKDIR /basedir/NFD_DEV_241114_A9/
+    setup_dbt latest_v5 && \
+    dbt-create -n "$NIGHTLY_TAG" && \
+    cd "$NIGHTLY_TAG" && \
+    source env.sh && \
+    dbt-build && \
+    dbt-workarea-env
+WORKDIR /basedir/$NIGHTLY_TAG/
```
AdrianDAlessandro, line 22: Just curious, what the particular reason for sometimes using "double quotes" around the env var and other times not?

AdrianDAlessandro (approved):

drunc_docker_service/Dockerfile
cc-a: The quoting is good practice in a shell script context in case of a space in a file/directory name. I believe it's not required (and possibly isn't valid syntax even) for `WORKDIR` as its a Docker directive.

cc-a:
> The fact that we are using as base image something labelled `nigthly-241114` has nothing to do with the actual nightly build we are using?

It does. Updating the nightly means creating a base image manually with it's own tag then updating the env variable in the Dockerfile.

PR #265, opened 2024-12-05
Use 1x1 config as default session

cc-a:
# Description

Configures a new default session (`1x1`). In terms of complexity this is somewhere between the current default (`2x3`) and the lightweight (`lr-session`) options. The advantage of this configuration over `lr-session` is that it has required arguments for the state transitions.

Based the branch off of `tree-table` for ease of testing. I've tried this on Linux a couple of times without issue. Please could you try this out (with the full set of FSM transitions) on the various different platforms - M1 Mac, Intel Mac and Windows. It will likely still not work on M1 Mac due to the known issue with the AVX instruction set emulation.

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


dalonsoa (approved):
Checked on Windows and Intel Mac - all good, and quite fluid in Mac, actually (but still better in Windows)!

PR #280, opened 2024-12-13
:construction_worker: Add a step for building the docs in the CI.

dalonsoa:
# Description

To make sure they actually build, in case we forget to run them locally. Probably more important when/if they become more complex.

I've opted for including a step within the `test` job rather than running it in its own job to reduce overheads, but can be easily extracted, if deemed more appropriate.

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


AdrianDAlessandro (approved):
LGTM! I like the approach of keeping it in the tests workflow since it's testing that the docs build correctly

.github/workflows/ci.yml
```diff
@@ -31,6 +31,9 @@ jobs:
       - name: Run tests
         run: poetry run pytest --cov-report=xml
 
+      - name: Test building the docs
+        run: poetry run mkdocs build --strict
```
alexdewar, line 35: I'd do things in this order:

```suggestion
      - name: Test building the docs
        run: poetry run mkdocs build --strict

      - name: Run tests
        if: always()
        run: poetry run pytest --cov-report=xml
```

As it stands, it'll run the tests and, if that fails, it won't check the docs. Also, if there are doc warnings, it won't upload coverage.

alexdewar (approved):
I'd tweak things so that tests are always run and docs are always checked, even if one of them fails. See my comment. Other than that, all good!

.github/workflows/ci.yml
cc-a: Agree with this suggestion though GitHub suggests using `${{ !cancelled() }}` rather than `always()`. https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#always

cc-a (approved):
Looks good.

.github/workflows/ci.yml
dalonsoa: Done, with Chris alternative.

Issue #301, opened 2025-01-10
Controller UI Layout


cc-a:
We should take pass at laying out the controller UI properly to make use of the available space. We should use a flex layout as for the process manager UI.

cvzbynek:
This seems like a job for me!

PR #306, opened 2025-01-14
FSM state diagram

### LINKED ISSUE #300: Consider adding a diagram of the FSM to the Controller UI
Suggestion from Alessandro:
image

Suggest starting with a mockup to run past the project team.

jamesturner246:
# Description

Add a flowchart representing the FSM state to the controller main page.

Fixes #300 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/controller/templates/controller/index.html
+++ b/controller/templates/controller/index.html
@@ -30,6 +30,8 @@
         href="https://cdn.jsdelivr.net/npm/@shoelace-style/shoelace@2.18.0/cdn/themes/light.css" />
   <script type="module"
           src="https://cdn.jsdelivr.net/npm/@shoelace-style/shoelace@2.18.0/cdn/shoelace-autoloader.js"></script>
+  <script type="module"
+          src="https://cdn.jsdelivr.net/npm/mermaid/dist/mermaid.min.js"></script>
 {% endblock extra_js %}
 {% block content %}
   <div class="container-fluid no-padding no-margin">
@@ -60,6 +62,24 @@
                  hx-trigger="load"
                  hx-target="#app_tree_summary"></div>
             <div id="app_tree_summary"></div>
+          </div>
+        </div>
+      </div>
+      <div class="fst-panel">
+        <div class="card shadow-sm rounded">
+          <div class="card-header bg-primary text-white rounded-top mb-3">
+            <h5>Application Diagram</h5>
+          </div>
+          <div class="card-body">
+            <div class="mermaid">
+              graph TD
+              A[Start] --> B{Decision?}
+              B -->|Yes| C[Task 1]
+              B -->|No| D[Task 2]
+              C --> E[End]
+              D --> E[End]
+            </div>
+            <script>mermaid.initialize({ startOnLoad: true });</script>
           </div>
         </div>
       </div>


controller/views/partials.py
```diff
@@ -27,11 +27,24 @@ def state_machine(request: HttpRequest) -> HttpResponse:
...
+
+    link = 0
+    flowchart = ""
+    for state, events in states.items():
+        for event, target in events.items():
+            flowchart += f"{state}({state}) -->|{event}| {target}({target})\n"
+            if state == current_state:
+                flowchart += f"style {state} fill:#93c54b\n"
+                flowchart += f"linkStyle {link} background-color:#93c54b\n"
+            link += 1
```
dalonsoa, line 43: I'd put this in a separate function, and here just:
```suggestion
    flowchart = make_flowchart(states, current_state)
```

controller/templates/controller/partials/state_machine.html
```diff
@@ -1,6 +1,42 @@
...
+      %% FSM transitions (edges).
+      linkStyle default background-color:#b5b3ae,stroke-width:2px
+
+      {{ flowchart }}
+    </pre>
+  </div>
+</div>
+<script>
+  // Initialize mermaid diagrams.
+  mermaid.initialize({
```
dalonsoa, line 26: I guess there's no way of doing this without JS, right?

controller/templates/controller/partials/state_machine.html
```diff
@@ -1,6 +1,42 @@
...
+<div class="row">
+  <div class="col-md-auto">
+    <form>
+      {% csrf_token %}
+      {% render_table table %}
+    </form>
+  </div>
+  <div class="col-md-auto">
+    <pre class="mermaid">
+      flowchart TD
```
dalonsoa, line 12: I wonder if it would look nicer LR instead of TD...

controller/templates/controller/partials/state_machine.html
```diff
@@ -1,6 +1,42 @@
...
+      edgeLabelBackground: "transparent", // HACK: custom edge label colour
+    },
+  });
+  // Refresh mermaid diagrams after htmx swap.
+  document.body.addEventListener("htmx:afterSwap", (event) => {
+      const target = event.detail.target;
+      if (target && target.querySelector(".mermaid")) {
+          mermaid.run({nodes: target.querySelectorAll(".mermaid")});
+      }
+  });
```
dalonsoa, line 41: I don't get this. Why does the diagram need to be manually refreshed? The whole block should be re-loaded and, presumably, a brand new diagram created with the new content.

controller/views/partials.py
dalonsoa: Appropriate tests would need to be added.

dalonsoa (changes requested):
I have a couple of comments/question, and a request. 

Having a proper diagram is nice, but currently, the look and feel and the position at the bottom of the page is not that great. I get this:

image

The diagram is tiny, hard to read and, for some reason, makes the column with the FSM stuff wider than it should, with extra blank space at the right for no reason because the diagram does not extend that far. A few suggestions would be:

- Try to make the diagram bigger, and maybe horizontal, and at the top, rather than the bottom.
- Whatever the orientation, make sure the table is as wide as the FSM block.
- Use dark blue font of the active state.
- For the transitions, use transparent/white background, dark blue fonts for the active ones and grey for the not active ones.

I am not sure if any of these suggestions would make it any better, but as it is now, it is not that readable.

jamesturner246:
> The diagram is tiny, hard to read and, for some reason, makes the column with the FSM stuff wider than it should, with extra blank space at the right for no reason because the diagram does not extend that far. A few suggestions would be:

Yeah, the diagram is supposed to be in a column to the right of the FSM table, but it doesn't look great for small screens.

However, I'm a bit hesitant to mess around with the layout, given @cvzbynek will be working on the controller page layout in #301. Perhaps we can review layout, and TD or LR in the PR once done?

controller/templates/controller/partials/state_machine.html
jamesturner246: The problem with mermaid is the diagrams are only rendered once, when the *main* document finishes loading. Content swaps with HTMX posts do not trigger another rendering, so it has to be this way.

controller/templates/controller/partials/state_machine.html
jamesturner246: See https://github.com/ImperialCollegeLondon/drunc_ui/pull/306#issuecomment-2604757006

dalonsoa:
I would not waste too much time with the layout, but I would definitely give it another pass. As it sits now, it's really neither nice nor useful. My screen is not small, but 4 columns side by side might be way too much for any sensible screen.

controller/templates/controller/partials/state_machine.html
jamesturner246: Afraid not.

controller/views/partials.py
jamesturner246: Done.

jamesturner246:
Okay, I've implemented your suggestions where possible @dalonsoa. Diagram is at top, and highlighted node font colour is dark blue. It may well be changed soon, but it hopefully looks alright for now.

jamesturner246:
Screenshot from 2025-01-21 18-54-59

PR #323, opened 2025-01-28
Moves the interfaces-related code to its own package

### LINKED ISSUE #274: Avoid cross reference between applications
> In the interest of not blocking this further, leave it as as and open an issue about it.

> I think the way we've been doing this is to put any common functionality between the apps into the `main` app (although I don't know if this actually is better for splitting things down the line)

_Originally posted by @AdrianDAlessandro in https://github.com/ImperialCollegeLondon/drunc_ui/pull/262#discussion_r1877777358_

At the moment, the controller app needs to import the process manager app in order to get the hostname information. That's not ideal. The best solution is unclear, but could consist on putting the `get_hostname` functionality in the `main` app.

dalonsoa:
# Description

The process and controller interfaces have been moved to their own root-level package - not a Django app - to avoid having cross-talks between applications. All future interfaces to drunc endpoints (eg. session manager) should be added here. 

There are no functional changes, just moving things around. Specifically:
- `process_manager_interface.py` and `controller_interface.py` have been moved to the new `interfaces` package
- Related test files have also been moved to `tests/interfaces`
- Import statements and mocks have been updated to reflect the new structure
- One function (`get_app_tree`) has been move to the `controller/app_tree.py` module to avoid `interfaces` importing the `controller` app. 

Fixes #274 

## Type of change

- [x] Breaking change (whatever its nature)

## Key checklist

## Further checks


dalonsoa:
@AdrianDAlessandro @cc-a I've added you as reviewers here as you pointed out the problem and suggested the solution, respectively.

jamesturner246 (approved):
Perfect. :+1:  Looks much nicer. 

`app_tree` is good where it is now; it doesn't do any RPC calls directly anyway. 

Will merge at the end of the day.

AdrianDAlessandro (approved):
This makes sense to me

PR #332, opened 2025-02-04
299 ers messages

### LINKED ISSUE #299: Messages in the ERS feed are not of type BroadcastMessage
Deployment testing with the Dune team highlighted that. The content of messages sent by Kafka contain a string encoded gRPC object. Messages for the [process manager use the BroadcastMessage type](https://github.com/ImperialCollegeLondon/drunc_ui/blob/9561eec1db5c9ccb9390595a56976ef31f476d64/main/management/commands/kafka_consumer.py#L67-L69) however this is not the case for controller messages on the ERS feed. We need to figure out the correct format and adjust our message handling logic. Hopefully we won't need to make any adjustments to our models to account for this.

dalonsoa:
# Description

Adds the handling of ERS messages by the Kafka consumer.

The assumptions here are that:
1. The format of the messages returned by `consummer.poll` is the same in both cases.
2. The specific change in formatting is within `message.value`, affecting mostly how to read the actual message and the severity.
3. For `time`, I've chosen to use `message.timestamp` in both cases, but for the `ERS` messages, we could also use `ic.final.time`, transforming from nanoseconds instead of miliseconds. I'm not sure if there's an advantage on either option. 

The ERS messaging handler cannot be tested in real life until:
1. The [repo containing the ERS message schema](https://github.com/DUNE-DAQ/ers) is converted into a Python package that can be added as a dependency and used in here.
2. We have a version of `drunc` actually broadcasting ERS messages that need handling.

So, for now, if we are happy with how the implementation looks, I'd suggest we merge it and come back when we can actually test this.

Fixes N/A

## Type of change

## Key checklist

## Further checks


jamesturner246 (approved):
LGTM :+1: This is fine, thanks.

PR #334, opened 2025-02-06
Adds the session manager frontend

### LINKED ISSUE #325: View and template for session manager
Create a view function which queries (currently fake) active session and available config data, from the functions defined in #320. 

Also create a template renders them as a django table, as below:

Image

We should try to keep the layout of the page as similarly formatted as we can to the other pages, allowing drunc devs to extend the page in future.

The active session columns are self-explanatory -- use the sess name and actor returned by the placeholder active sess function.

The available configuration column, unlike in the sketch above, should be two columns -- one with the config file and the other with config id, as returned by the placeholder get available config function.

dalonsoa:
# Description

Adds a new app with the relevant tables, views, templates and urls, showing the two required tables with placeholder data.

I've also changed the `get_sessions` function to return a hardcoded set of sessions since, with the auto-refresh of the page, it was populating the able with random entries every second. 

https://github.com/user-attachments/assets/11d91742-6f99-485d-aea0-cf9d50949700

Fixes #325

## Type of change

## Key checklist

## Further checks


dalonsoa:
@jamesturner246 I've updated this to create a brand new app, as agreed.

jamesturner246 (changes requested):
Looks nice and neat. Hi Gandalf! :+1: 

Just one thing about moving the sess man interface to its own module and that's all good.

jamesturner246 (approved):
Perfect :+1:

PR #336, opened 2025-02-07
Adds ERS dependency

### LINKED ISSUE #299: Messages in the ERS feed are not of type BroadcastMessage

dalonsoa:
# Description

I've opted to directly build the dependency and add it to the repository - following Pierre instructions here - as it facilitates both local development and testing in the QA system, but ideally we would like it to be a standalone dependency like `druncschema` installable like any other package.

Adding this still does not allow us to try the new code, as we have nothing at hand broadcasting this kind of messages, but at least now the code is complete and, bugs aside, should work. 

Fixes #299 

## Type of change

## Key checklist

## Further checks


jamesturner246:
Thanks @dalonsoa. Looking now.

Regarding the `druncschema` and other dependencies, I'm looking into solving that very issue now. `druncschema` already is its own package, we just need a way of injecting our own `druncschema` and `drunc` copies into the provided docker images and we are good. Next sprint.

main/management/commands/kafka_consumer.py
```diff
@@ -70,23 +69,12 @@ def from_ers_message(message: Any) -> DruncMessage:  # type: ignore [misc]
...
-            timestamp=time,
-            message="ERS message could not be processed.",
-            severity="ERROR",
-        )
-
     ic = IssueChain()
     ic.ParseFromString(message.value)
 
-    # TODO: Check if the value of this is valid.
-    severity = ic.final.severity.upper()
```
jamesturner246, line 89: Is both `FATAL` and `CRITICAL` returned from ERS? I wonder if it's worth just altering our model to `FATAL`, if that's what is used now.

jamesturner246 (approved):
LGTM. :+1: 

Just one question: wondering if it's worth updating the model if one of the statuses in unused.,

main/management/commands/kafka_consumer.py
dalonsoa: ERS only return FATAL, no critical. But, in principle, we have CRITICAL severity in normal Kafka messages (See line 27 above). We could change that one to FATAL, change in the model CRITICAL by FATAL, and everything will be simpler and consistent. I would say that is the way to go.

jamesturner246: That sounds good to me.

dalonsoa: OK, so I've made the changes in [d5e75aa](https://github.com/ImperialCollegeLondon/drunc_ui/pull/336/commits/d5e75aa7e1fb1902b2306fea05ce3bdb67d04c3e)

PR #359, opened 2025-02-20
Install local deps into container

### LINKED ISSUE #358: `drunc`/`druncschema` devel container

jamesturner246:
# Description

Install local versions of `drunc` and `druncschema` into containers in editable mode, to enable developing and testing with the UI. 

It seems like docker really really doesn't like this, so we should probably revisit. I don't know a cleaner way.

The repos are searched at `DRUNC_LOCAL_DEPS`, so make sure they're cloned to the right place.
This defaults to the current directory from which `install_local_deps.sh` is run, if `DRUNC_LOCAL_DEPS` is unset.

Then, with the compose up, run the following to install the repos editable inside the containers:

```sh
./scripts/install_local_deps.sh
```

Fixes #358.

## Type of change

## Key checklist

## Further checks


dalonsoa (changes requested):
I've added the Powershell version of the script and tested it works as intended, letting me use the locally installed version of the repositories. 

The only change required is to add instructions to the developer documentation, so other's can make this work themselves, if needed.

jamesturner246:
Done.

docs/development.md
```diff
@@ -184,4 +184,29 @@ from the database after a configurable retention period.
...
+# Defaults to current directory if unset.
+export DRUNC_LOCAL_DEPS=/path/containing/dep/repos
+
+docker compose --profile drunc up -d
+./scripts/insttall_local_deps.sh
+```
+
+An equivalent powershell script is also provided for Windows:
+
+```powershell
```
dalonsoa, line 208: For completeness:

```suggestion
```powershell
[System.Environment]::SetEnvironmentVariable('DRUNC_LOCAL_DEPS', '/path/containing/dep/repos')

docker compose --profile drunc up -d
```

dalonsoa (approved):

PR #370, opened 2025-02-27
Starts the Session Manager in the `drunc` container

### LINKED ISSUE #367: Start session manager in docker compose
With placeholder session manager implemented in drunc backend, and docker compose changed to allow local versions of `drunc` and `druncschema` to be installed, we can now start an instance of session manager in the compose.

Add commands in `drunc_docker_service/Dockerfile`.

dalonsoa:
# Description

And opens the relevant ports, I think. 

While the process manager has two distinct commands, `drunc-process-manager-shell` and `drunc-process-manager`, the session manager only has one, and it runs without arguments, so it's unclear if they should be run in the same way. For example, if `drunc-process-manager` is blocking, then `drunc-session-manager` will not run in the docker compose file...

Open as draft to get feedback.

Fixes #367 

## Type of change

## Key checklist

## Further checks


drunc_docker_service/Dockerfile
```diff
@@ -36,3 +36,6 @@ ADD wibeth_output_all_zeros.bin /cvmfs/dunedaq.opensciencegrid.org/assets/files/
 
 ADD boot_test_session.sh /
 RUN chmod a+x /boot_test_session.sh
+
+ADD boot_session_manager.sh /
+RUN chmod a+x /boot_session_manager.sh
```
jamesturner246, line 41: Don't think we need this.

docker-compose.yml
```diff
@@ -26,9 +27,11 @@ services:
       - -c
       - |
         /usr/sbin/sshd &&
-        drunc-process-manager --log-level debug /process-manager-kafka.json 10054
+        drunc-process-manager --log-level debug /process-manager-kafka.json 10054 &&
```
jamesturner246, line 30: Can make `drunc-process-manager` fork to background with a single `&`, but see main comment.
```suggestion
        drunc-process-manager --log-level debug /process-manager-kafka.json 10054 &
```

jamesturner246 (commented):
Can make the PM non-blocking with `&` instead of `&&`, but I think the way to go here is probably to have a separate container for session manager. 

I know it's another whole container with the full drunc stack inside -- in an ideal world, the session manager and process manager inside `drunc` would be separated out from core `drunc` code into their own repos, but this is what we have for now.

dalonsoa:
Why would be better separate containers? I don't mind either way - although two containers is not ideal - but just trying to understand the drawbacks of having both in the same container.

jamesturner246:
For starters, better encapsulation. If one fails, it doesn't wreck the whole container. Also debugging is a fair bit easier, and we don't have any of these blocking problems you described above.

It seems to be so [by convention](https://www.baeldung.com/ops/one-process-per-container), as well.

dalonsoa:
@jamesturner246 Ok, so this implements your hack suggestion and seems to work fine, so until the `session_manager` is in the nightly build, is would do. Please, have a look and merge - or not :)

jamesturner246 (approved):
LGTM. Thanks

PR #393, opened 2025-03-20
Creates new nightly for the DUNEDAQ development envionment

### LINKED ISSUE #383: Create new nightly

dalonsoa:
# Description

Uses a new nightly build for the DUNEDAQ environment. A bunch of variables here and there has been changed to make use of the new architecture of separate services for the process manager and the session manager, as well as various changes in how the drunc backend works. All the existing functionality in `drunc_ui`, as well as the simple testing scripts have been checked to work as planned. 

The new nightly build for us is now available in https://github.com/ImperialCollegeLondon/drunc_ui/pkgs/container/dunedaq_dev_environment

Fixes #383 

## Type of change

## Key checklist

## Further checks


jamesturner246 (approved):
Looks sensible. I don't fullyunderstand SESSION vs SESSION_NAME in PM shell, and boot, CONFIG etc, but okay.

dalonsoa:
Essentially, they have changed what is needed to boot a session. SESSION is now the old session id - `1x1-whatever` - while SESSION_NAME is a human-readable, but otherwise arbitrary name. The trick is that you need to use that same name when dealing with the session, so while this can be set via environment variables, I've made it default to SESSION in our setup.

PR #433, opened 2025-05-01
Deployment container fixup

### LINKED ISSUE #430: Deployment example not working
Diagnose deployment example no longer working, and fix/update docs as necessary.

jamesturner246:
# Description

General fixup and testing of the deployment compose with the latest software. The main app container in the compose now has its own `home` space, allowing it to write its own files (like static site files) without volume mounting and permissions jankiness. 

This should also make using custom versions/tags of the drunc dependencies easier, since we can now just clone them directly into this home space in the container.

Deployment documentation is checked over and modernised.

Fixes #430 

## Type of change

- [x] Documentation (non-breaking change that adds or improves the documentation)

## Key checklist

## Further checks


dalonsoa (changes requested):
For this to work locally, you need to change the `docker-compose.yml` file in the root directory, as well. In particular, lines 13 and 14 need to change to:

```
      - .:/home/dune_user/app
      - db:/home/dune_user/app/db
```

Otherwise, launching the `app` fails with:

```
app-1  | python: can't open file '/home/dune_user/app/manage.py': [Errno 2] No such file or directory
```

As you are in a different directory that where you mounted the local directory.

jamesturner246:
Oops. Should be fixed now.

dalonsoa (approved):
I've tested it and works fine 👍

PR #468, opened 2025-06-17
:construction_worker: Add schedule test with Python 3.11

### LINKED ISSUE #278: Run python 3.11 in CI
We've swapped to developing primarily against Python 3.10 but it's a good idea to future proof and make sure we also worked with 3.11. This implies we should have CI run against 3.11 but in line with our group policy of minimising actions usage this should be less frequent then on every pull request. We could run on a schedule or adopt a more formal release model with associated CI.

dalonsoa:
# Description

Create a separate workflow that runs the tests using Python 3.11 occasionally, checking that all works - or will work - in future versions of Python. 

Fixes # (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,8 +23,7 @@ jobs:
       matrix:
         os:
           - ubuntu-latest
-        python-version:
-          - ${{ inputs.python_version }}
+        python-version: ${{ inputs.python_version }}
 
     steps:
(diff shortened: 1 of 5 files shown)


dalonsoa:
Will reopen when/if needed, as the plans are to bump the whole of the DUNEDAQ ecosystem to a newer python version by September or so.

PR #469, opened 2025-06-18
Update the JS dependencies and add instructions

### LINKED ISSUE #445: Remove JS code and include the appropriate links to use it by the client

dalonsoa:
# Description

Adds a README explaining the built-in JS depdencencies, their current versions and how to update them, should it be necessary. It also updates them to the latests versions. As far as I can tell, they all work fine (no major version updates)

Fixes #445

## Type of change

- [x] Documentation (non-breaking change that adds or improves the documentation)

## Key checklist

## Further checks


dalonsoa:
The security aspect mentioned in the post is one of the main reasons, specially considering that we are talking about CERN and we don't want the control software to be down because of some JS libraries stored somewhere are not available at run time. But also a bit on keeping JS dependencies clear and explained somewhere separated from where they are used.

Issue #513, opened 2025-08-07
No static files available if DEBUG=False


dalonsoa:
Not relevant, for now, but will become relevant as soon as we try to run DRUNC_UI in a production environment, where `DEBUG=False`. As it is now, when DEBUG is False, we get for the process manager page:

<img width="606" height="371" alt="Image" src="https://github.com/user-attachments/assets/f0558dd0-ee9e-4df0-a2ff-faedcee4f780" />

Other pages will have similar look and feel and many things are not working since JS files we are shipping alongside `drunc_ui` are simply not available. 

More information in: https://docs.djangoproject.com/en/5.1/howto/static-files/

dalonsoa:
This is not really a bug in the code, but rather on how it is deployed. In other words, it depends on how the reverse proxy serving the web app is configured to serve also static files.

In development, where we use `DEBUG=True`, we do not need a reverse proxy. And in production, how things are done very much depend on the deployment strategy. In our standard setup with `docker compose`, we often use `Caddy` in a way similar to how it is described [in this post and related comments](https://stackoverflow.com/questions/77439324/caddy-how-to-serve-static-files-from-django). Other common tool is using `nginx`, as described in detail [in this post](https://medium.com/django-unleashed/serving-static-and-media-files-using-nginx-in-django-a4a125af95d).

As the plan seems to be using k8 for deployment, that will require a different configuration to ensure that static files are served, but the spirit will be the same that the posts above using docker compose. 

I leave this issue open as reference when designing the deployment strategy, but there is nothing that, at the moment, needs changing in the code. 

@PawelPlesniak , @miruuna , @Aurashk , @jamesturner246

dalonsoa:
Closing, as it has to do with deployment, not code.

PR #514, opened 2025-08-07
:sparkles: Disable boot button if debug=False

### LINKED ISSUE #297: Disable dummy boot functionality in production
The process manager UI currently has a "Boot" button. This invokes the dummy_boot command of the process manager. This is useful for testing in development but shouldn't be present in production. We need to:

- Use the DEBUG setting to control whether the feature is present.
- If DEBUG=False hide the "Boot" button in the process manager UI.
- If DEBUG=False do not include the `boot_process` view in `urls.py`

dalonsoa:
# Description

Remove the "Boot" button on the `Process Manager` page when the tool runs in production - i.e. when `DEBUG=False` in the settings. 

I've tested this works and indeed the button is no longer there and the URL for booting sessions is not accessible, however that has revealed another issue (described in #513 )

Fixes #297

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/process_manager/templates/process_manager/index.html
+++ b/process_manager/templates/process_manager/index.html
@@ -31,8 +31,11 @@
             <form method="post" action="{% url 'process_manager:process_action' %}">
               {% csrf_token %}
               <div class="d-flex justify-content-between mb-3">
-                <a href="{% url 'process_manager:boot_process' %}"
-                   class="btn btn-primary w-100 me-2">Boot</a>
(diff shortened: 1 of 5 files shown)


jamesturner246 (approved):
Looks good, but strikes me as something that will require tests. e.g. to check the url_path/button exist when `DEBUG==False`. 

Up to you if you want to put in this PR or later, so I'll approve.

PR #532, opened 2025-08-28
JCF: add the same workflows as are currently available in drunc

### LINKED ISSUE #531: Add standard set of Workflows
Now that `drunc_ui` has been added to the `DUNE-DAQ` organization, we need to give it an appropriate set of workflows.

jcfreeman2:
# Description

This adds the `drunc` Workflows now that as of this morning `drunc_ui` has been added to the GitHub organization. 

Fixes #531

## Type of change

## Key checklist
N/A, this consists of Workflow additions rather than code changes. 

## Further checks

### DIFF
--- a/.github/workflows/run_mqst.yml
+++ b/.github/workflows/run_mqst.yml
@@ -1,0 +1,57 @@
+name: Run MQST
+
+on:
+  never-ever:
+
+jobs:
+  install_and_test:
+
+    runs-on: ubuntu-latest
+
+    defaults:
+      run:
+        shell: bash
+
+    container:
+      image: ghcr.io/dune-daq/nightly-release-alma9:development_v5
+
+    steps:
+      - name: Install SSH and run the daemon
+        run: |
+          dnf install openssh-server
+          dnf install openssh-clients
+          systemctl start sshd
+          systemctl enable sshd
+          ssh localhost echo "SSH is working"
+
+      - name: Set up dev nightly
+        run: |
+          mkdir nightly
+          cd nightly
+          source /cvmfs/dunedaq.opensciencegrid.org/setup_dunedaq.sh
+          setup_dbt latest_v5
+          dbt-create -n last_fddaq
+          cd NFD_DEV*
+          source env.sh
+          dbt-build
+          dbt-workarea-env
+
+      - name: Install druncschema and drunc
+        run: |
+          cd nightly/NFD_DEV* || true
+          source env.sh
+          python -m pip install --upgrade pip
+          git clone https://github.com/DUNE-DAQ/druncschema.git -b ${{ github.head_ref }} ||  git clone https://github.com/DUNE-DAQ/druncschema.git
+          git clone https://github.com/DUNE-DAQ/drunc.git -b ${{ github.head_ref }}
+          cd druncschema
+          pip install .
+          cd -
+          cd drunc
+          pip install .
+          echo "DRUNC_DIR=$(pwd)" >> $GITHUB_ENV
+
+      - name: Test with MQST
+        run: |
+          cd nightly/NFD_DEV* || true
+          source env.sh
+          daqsystemtest_integtest_bundle.sh -l 0
--- a/.github/workflows/run_pytest.yml
+++ b/.github/workflows/run_pytest.yml
@@ -1,0 +1,106 @@
+name: Run pytest
+
+on:
+  pull_request:
+
+jobs:
+  install_and_test:
+
+    runs-on: ubuntu-latest
+
+    defaults:
+      run:
+        shell: bash
+
+    container:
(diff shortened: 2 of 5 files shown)


.github/workflows/track_new_issues.yml
dalonsoa, line 1: I know what this does - I think - but not why you need it.

GitHub Projects have built-in capability of automatically track new issues (among other things), so I'm not sure what this custom workflow adds to the table. Unless you need some specific subtlety, I think it would be better to stick to the out of the box solution. See video below how it is setup for the other project board we have:

https://github.com/user-attachments/assets/2a58588f-9eab-4263-9934-796d53401090

.github/workflows/run_pytest.yml
dalonsoa, line 1: As many of the dependencies of the GUI are currently missing in the environment, this workflow will fail. Not an issue in itself, but it will be annoying to have workflows failing until we fix that.

.github/workflows/track_new_prs.yml
dalonsoa, line 1: Same with this one. You can also track PRs with the built-in capabilities of GitHub Projects.

.github/workflows/run_pytest.yml
```diff
@@ -0,0 +1,106 @@
...
+          cd NFD_DEV*
+          source env.sh
+          dbt-build
+          dbt-workarea-env
+
+      - name: Install druncschema and drunc
+        if: success()
+        run: |
+          cd nightly/NFD_DEV* || true
+          source env.sh
```
dalonsoa, line 41: Setting this environment based on the last nightly is used in a couple of repositories, and it is relatively complex. I would suggest to create a separate action that can be reused by any repo that needs it such that if changes are needed, they can be done in a single place.

.github/workflows/run_mqst.yml
dalonsoa, line 1: @PawelPlesniak , could you give a brief summary of what this workflow does?

.github/workflows/run_mqst.yml
```diff
@@ -0,0 +1,57 @@
+name: Run MQST
+
+on:
+  never-ever:
```
dalonsoa, line 4: This is confusing...

dalonsoa (commented):
I've left a few comments, maybe not so much things to change in this PR but rather things to consider to simplify things in the future.

@PawelPlesniak , it will be good to have your input in some of these.

.github/workflows/run_mqst.yml
PawelPlesniak: The MSQT is the minimal high level integration test that we want to run, but is being held back by the fact that I haven't finished the separate type of process manasger (subprocess). The intention is for this to be run on merge in the future, but for now it cannot run due to the permissions of running ssh commands in CI tooling

PawelPlesniak:
Added my (very limited) comments
For the discussion of how we handle PR and Issue tracking, I leave this with @jcfreeman2

jcfreeman2:
One thing I think this PR experience has underscored is that along with the more complex aspects of standardizing Python package Workflows (e.g., Workflows needing certain dependencies (*) or even requiring updates to drunc for some of them to run) there's also low hanging fruit. E.g., for the C++ packages we already centralize the treatment of adding Issues and PRs - e.g., if you look at 
https://github.com/DUNE-DAQ/appfwk/blob/develop/.github/workflows/track_new_issues.yml
you see that it calls a centralized Workflow,
https://github.com/DUNE-DAQ/.github/blob/develop/.github/workflows/track_new_issues.yml

I've mentioned this to Andrew and he agrees that we should standardize `track_new_issues.yml` and `track_new_PRs.yml` for the Python packages (`drunc`, `runconf-ui`, etc.).  While interacting with the GitHub GUI is quite powerful and convenient, one advantage of having the `*.yml` files is that it leaves a written, version-controlled record of how we're handling the tracking, plus it leaves us flexibility to make nonstandard changes using `graphql`.

I'd also add that Diego's comment about the testing Workflow "I would suggest to create a separate action" is correct on the merits and Software Coordination should [centralize it in the `.github` repository](https://github.com/DUNE-DAQ/.github/) as it's done with other workflows. 

Final point about the (Draft) PR itself: rather than merge the branch which pulls in all the Workflows as-is from `drunc`, let's think of this as having created a useful discussion, figure out what Workflows for Python packages should look like, and then provide them to `drunc_ui` as we'll provide them to `drunc`, `github-ui`, etc. 

(*) Pawel and I will talk about this at Software Integration and Testing on Sep. 16th

jcfreeman2:
Closing this for reasons covered in the line concering how we should "figure out what Workflows for Python packages should look like, and then provide them to drunc_ui as we'll provide them to drunc, github-ui, etc."

PawelPlesniak:
Just a quick note - having the template issues and PRs in the central .github repository would make it easier to use the defined templates.

PR #551, opened 2025-11-04
Handle errors in Controller

### LINKED ISSUE #298: UI behaviour when controller is unreachable
Currently if the controller UI is opened and the root controller is unreachable (e.g. because it is not running) then the Finite State Machine and Application Tree boxes on the page simply do not populate and the user receives no feedback. Under the hood the connectivity server raises an exception when it is unable to find the root controller and the htmx requests used to populate the page elements get a 500 response.

A simple improvement we could make here would be to handle the error raised by the connectivity server and return a generic "Unable to contact root controller" message.

We already have some infrastructure to help with this in the form of the `main/views/utils.py::handle_errors` decorator. This could be fairly straightforwardly enhanced to allow a customisable error message and applied to the relevant view functions in the controller.

dalonsoa:
# Description

Produces a somewhat meaningful error message for the user when the controller, the process manager and the session manager cannot be connected for any reason. It expands the existing error message, logging the full stack but reporting to the end user a simpler version of the error - for security and because they don't need more. 

[Screencast from 2025-11-06 09-15-45.webm](https://github.com/user-attachments/assets/71ccc42e-4da0-4ad1-9818-2fe81445a348)

Fixes #298 

## Type of change

## Key checklist

## Further checks


docker-compose.yml
```diff
@@ -62,7 +62,7 @@ services:
         condition: service_healthy
 
   kafka:
-    image: bitnami/kafka:latest
+    image: ubuntu/kafka:latest
```
dalonsoa: Wrong. Removing this change. 

```suggestion
    image: bitnami/kafka:latest
```

main/views/utils.py
```diff
@@ -11,22 +12,29 @@
...
+    def decorator(view_func: ViewType) -> ViewType:
+        @wraps(view_func)
+        def wrapped_view(request: HttpRequest, *args, **kwargs) -> HttpResponse:  # type: ignore
+            try:
+                return view_func(request, *args, **kwargs)
+            except Exception as e:
+                logger = logging.getLogger("django")
+                logger.exception(e)
+                context = {"error_message": error_message}
+                return render(request, "main/error_message.html", context=context)
```
miruuna: It might be useful to also return a status code in here.

interfaces/controller_interface.py
```diff
@@ -29,7 +29,7 @@ def get_controller_uri() -> str:
         str: The URI of the root controller.
     """
     csc = ConnectivityServiceClient(settings.CSC_SESSION, settings.CSC_URL)
-    uris = csc.resolve("root-controller_control", "RunControlMessage")
+    uris = csc.resolve("root-controller_control", "RunControlMessage", ntries=10)
```
miruuna: I wonder if we could reduce the number of tries (perhaps 5 tries), to reduce the time it takes for the message to appear, and therefore improve the user experience a little bit?

jamesturner246 (changes requested):
Thanks @dalonsoa. Looks good. I'd agree with @miruuna though. Showing error code/exception message, and making the error appear a bit sooner with fewer retries. The user can always refresh if they want to keep trying beyond that.

Aurashk:
I'm not 100% sure if this will work but it might be nice to render a loading message in the html initially while you are waiting to fetch the result which will get overwritten when the request comes back. I think in https://github.com/DUNE-DAQ/drunc_ui/blob/218ee13fbbb8ee040f02c8c958ca138bbb4f3dbd/controller/templates/controller/app_tree_view.html#L17 you would just need to replace 
```
<div class="card-body">
  <div hx-get="{% url 'controller:app_tree_table' %}"
       hx-trigger="load"
       hx-target="#app-tree-table"></div>
  <div id="app-tree-table"></div>
</div>
 ``` 
 
 with something like

```
<div class="card-body">
  <div hx-get="{% url 'controller:app_tree_table' %}"
       hx-trigger="load"
       hx-target="#app-tree-table"></div>
  <div id="app-tree-table">
    <p class="mt-2 text-muted">Waiting for connectivity service...</p>
   </div>
</div>
```

Or make it more fancy and have `<p class="mt-2 text-muted">Waiting for connectivity service...</p>` be a loading spinner or something.

Otherwise looks good!

dalonsoa:
Many thanks for your comments. I've updated the PR description above with the new changes (and a new video).

Aurashk (approved):
Looking nice!

jamesturner246 (approved):
Nice. Thanks @dalonsoa. 

I'm not sure if we have rights, literal and figuratively, to merge any more so I'll leave it open until we can ask. @PawelPlesniak ?

PawelPlesniak:
You are welcome to merge. With the release deadline upcoming this Wednesday, I have not had time to test this with k8s, but if you are happy to merge then I am too. I will be testing this next week, sorry about the continuous delays with this

